PR #9, opened 2015-06-28
Setup CI and badges

AdrieanKhisbe:
Hello @bastibe ! :)

That's a nice package that you started building! 

I set up for you the test and continous integration so you can more easily start to write test on the go.

so far just scaffold test, but a least you're protected from syntax error:
here is a status build https://travis-ci.org/AdrieanKhisbe/annotate.el/builds/68666866

If you merge it, you'll have to active travis and coveralls with your github account.

about Cask, you'll find most information there: http://cask.readthedocs.org/en/latest/
(to install: http://cask.readthedocs.org/en/latest/guide/installation.html, (it's in brew), then just run `cask install` in your folder and you good to go! :))


bastibe:
Thank you! 

This would pull in a lot of dependencies, some of which don't seem to add much value to the project.  
Could you explain to me why cask would be desirable?  
Also, I don't see the point of the Makefile. Why would I need that?

I'll definitely build a test suite at some point, but I'll probably use _either_ ecukes _or_ ert, not both.

AdrieanKhisbe:
The dependancies are only for the development dependancy! :)
Nothing changed for the package, it's a fully iso PR (in term of features)!

Cask is there for standardise emacs package management, and easily retrieve dependancy for development. You'll find the motivation here. https://cask.github.io/why-cask.html

and the Makefile is there both for use and documentation: http://rejeep.github.io/emacs/cask/ert/ert-runner/ert-async/ecukes/testing/travis/2014/01/09/various-testing-tools-in-emacs.html

About Ecukes and Ert, both have they use some tests would be easier to do or one or another, so with the config for both, will be more easy to throw the first ecukes/ert test.

bastibe:
Upon researching this some more, I don't think that Cask would have any tangible benefit for me. 

This is a very simple and small project, with just one source file, a readme, and a license. It is installable through melpa/marmalade, or a simple `(require 'annotate)`. As long as it stays this simple, I don't have the problems Cask solves. Sorry.

A test bench would be nice, though. And once I have tests, Travis.ci would be nice. I'll refer back to this PR once I have a test bench. At the moment, I don't want to clutter the project directory with the stub files, though.

Again, thank you for your work, and I will refer back to it once the need arises, but at the moment, I don't want to include it into annotate.el yet.

AdrieanKhisbe:
Cask is to help the developer(s)
As a user perspective, nothing changed :)
As easy to install, and maybe less bugs.

All the other stuff I set up is too help you do test, have them run for most
of emacs version with coverage. Any of these files are included in the package made by melpa!
So don't hesitate to reuse it when you'll start doing your tests suit!

PR #25, opened 2016-04-11
Add support for local setting of the annotate-file.

matbra:
Hi @bastibe !

Thanks for this neat package. I'm using it a lot to make notes in org mode files and so far it works pretty great.

However, I like to keep my "projects" (mostly org mode files with related source code in their own directory) separated. Therefore, I tried setting `annotate-file` as a [file variable](https://www.gnu.org/software/emacs/manual/html_node/emacs/Specifying-File-Variables.html#Specifying-File-Variables). With the original implementation of `(annotate-load-annotation-data)` this doesn't work as you are creating a temporary buffer in which `annotate-file` still contains the global value (e.g. `~/.annotations`). Hence, I made a small change in your code rather soon after I had started using the package some months ago. As I'm not very familiar with Elisp I am not sure whether this is a proper solution - most probably there are better ways.

Anyways, I thought this might be a good opportunity to send my first pull request... :-) If I broke any convention, please let me know...

Do you think this could make sense - or am I getting something fundamentally wrong here?

Cheers

Matthias

### DIFF
--- a/annotate.el
+++ b/annotate.el
@@ -627,7 +627,9 @@ an overlay and it's annotation."
 
 (defun annotate-load-annotation-data ()
   "Read and return saved annotations."
+  (setq local-annotate-file annotate-file)
   (with-temp-buffer
+    (setq-local annotate-file local-annotate-file)
     (when (file-exists-p annotate-file)
       (insert-file-contents annotate-file))
     (goto-char (point-max))
--- a/annotate.el
+++ b/annotate.el
@@ -627,15 +627,16 @@ an overlay and it's annotation."
 
 (defun annotate-load-annotation-data ()
   "Read and return saved annotations."
-  (setq local-annotate-file annotate-file)
-  (with-temp-buffer
-    (setq-local annotate-file local-annotate-file)
-    (when (file-exists-p annotate-file)
-      (insert-file-contents annotate-file))
-    (goto-char (point-max))
-    (cond ((= (point) 1) nil)
-          (t (goto-char (point-min))
-             (read (current-buffer))))))
+  ;; use the buffer-local value
+  ;; (the global value was possibly overwritten)
+  (let ((local-annotate-file annotate-file))
+    (with-temp-buffer
+      (when (file-exists-p local-annotate-file)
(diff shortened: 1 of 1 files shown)


annotate.el
```diff
@@ -627,7 +627,9 @@ an overlay and it's annotation."
 
 (defun annotate-load-annotation-data ()
   "Read and return saved annotations."
+  (setq local-annotate-file annotate-file)
```
bastibe: You should probably add a comment as to why you are copying the variable here. Also, this should probably be a `let` instead of a `setq`.

annotate.el
```diff
@@ -627,7 +627,9 @@ an overlay and it's annotation."
...
   (with-temp-buffer
+    (setq-local annotate-file local-annotate-file)
```
bastibe: It's probably nicer to replace the `annotate-file` in the next two lines with the new `local-annotate-file` instead of `setq-local`ing the variable.

bastibe:
That's a very good idea!

It should also be possible to `(setq annotate-file ".annotations")` globally, which should place the `.annotations` file in the current directory.

annotate.el
matbra: This makes sense. I changed it in 135fe586743ad4759d4bcfe53bbb343953c2e841.

annotate.el
matbra: I added a comment and probably should read a bit more about Lisp variables. All changes are in 135fe586743ad4759d4bcfe53bbb343953c2e841.

matbra:
You are completely right: globally setting `annotate-file` to e.g. `".annotations"` would surely have been the straightforward way to solve my problem.

Anyway, maybe supporting file-local variables is of use now and then...

bastibe:
Do you still need the changes from your pull request, or is setting `annotate-file` a better solution?

matbra:
Actually, I'm perfectly fine with setting `annotate-file`. It's okay for me to close this pull request. Sorry for making a fuss about this.

Cheers

Matthias

bastibe:
No need to be sorry, this is what pull requests are for!

PR #41, opened 2019-05-18
use `locate-user-emacs-file` to specify default save file

conao3:
Hi! This package is awesome!
I found small issue as Elisp package, I fix it.

---
It is recommended that Elisp packages save files in user-emacs-directory. The default value for it is `~/.emacs.d/`, and it has been changed by the user,
`locate-user-emacs-file` is returns path in consideration of the change.

And `locate-user-emacs-file` returns previous path if file exist in HOME, this change is not destructive.

cf: http://emacs.rubikitch.com/new-files/


bastibe:
Really cool, thank you!

Issue #6, opened 2015-06-15
Detect file changes and update annotations accordingly


bastibe:
Save a file hash next to the file name to detect file changes. Warn the user about file changes when annotations are loaded.

Possibly save each annotation's line to it so that it can be found if the text positions don't match any more.

cage2:
Hi!

Can this issue be closed after #47 is going to be merged?

Bye!
C.

PR #43, opened 2019-09-24
added file checksum in annotation metadata

cage2:
Hello!

With these commits i tried to address #6. I have added am md5 hash to each annotated file. When  annotate minor mode is loaded a check between the hash stored in the metadata (`.emacs.d/annotations`) and the current buffer is performed. If the two checksum are not equals a warning text is displayed.
When a file is saved an md5 sum is calculated for current buffer and stored in the metadata file.

Also added a 'blacklist' of major modes that prevents annotate to be loaded.

Hope i understood #6 correctly, though! :-)

Please consider merging.

bye!
C.

### DIFF
--- a/annotate.el
+++ b/annotate.el
@@ -291,8 +291,8 @@ major mode."
      (string-width (annotate-actual-comment-end))))
 
 (defun annotate-wrap-in-comment (&rest strings)
-  "Put comment  markers at  the start and  (if does  makes sense)
-ends  of   a  string.   See:   annotate-actual-comment-start  and
+  "Put comment  markers at  the start and  (if it  makes sense)
+end  of   a  string.   See:   annotate-actual-comment-start  and
 annotate-actual-comment-end"
   (apply #'concat (append (list (annotate-actual-comment-start))
                           strings
@@ -731,10 +731,14 @@ an overlay and it's annotation."
   (format "-%i,%i +%i,%i" start-line diff-size start-line diff-size)))
 
 (defun annotate-checksum-from-dump (record)
+  "Get the checksum field from  an annotation list loaded from a
+file."
   (and (> (length record) 2)
        (nth 2 record)))
 
 (defun annotate-annotations-from-dump (record)
+  "Get the annotations field from  an annotation list loaded from a
+file."
   (nth 1 record))
 
 (defun annotate-load-annotation-old-format ()
--- a/annotate.el
+++ b/annotate.el
@@ -126,7 +126,7 @@ less than this size (in characters)"
           "from the last time the annotations were saved.\n"
           "Chances are that they will not be displayed correctly")
   "The message to  warn the user that file has  been modified and
-  annototatnions position could be outdated")
+  annotations positions could be outdated")
 
 (defcustom annotate-blacklist-major-mode '(org-mode)
   "Prevent  loading of  annotate-mode  When  the visited  file's
@@ -135,6 +135,8 @@ major mode is a member of this list (space separated entries)."
   :group 'annotate)
 
 (defun annotate-initialize-maybe ()
+  "Initialize annotate mode only if buffer's major mode is not in the blacklist (see:
+'annotate-blacklist-major-mode'"
   (let ((annotate-allowed-p (with-current-buffer (current-buffer)
                               (not (cl-member major-mode annotate-blacklist-major-mode)))))
     (cond
@@ -147,12 +149,16 @@ major mode is a member of this list (space separated entries)."
      (annotate-shutdown)))))
 
 (cl-defun annotate-buffer-checksum (&optional (object (current-buffer)))
+  "Calculate  an   hash  for   the  buffer  'object',   skip  the
+  calculation     if     the     buffer    is     bigger     than
+  'annotate-maximum-size-checksum' (units are character)."
   (if (< (buffer-size)
          annotate-maximum-size-checksum)
       (md5 object)
     nil))
 
 (cl-defmacro annotate-with-inhibit-modification-hooks (&rest body)
+  "Wrap 'body' in a block with modification-hooks inhibited."
   `(unwind-protect
        (progn
          (setf inhibit-modification-hooks t)
@@ -180,11 +186,13 @@ major mode is a member of this list (space separated entries)."
                                   (1 (annotate--change-guard))))))
 
 (defun annotate-overlay-filled-p (overlay)
+  "Does this overlay contains an 'annotation' property?"
   (and overlay
        (overlayp overlay)
        (overlay-get overlay 'annotation)))
 
 (defun annotationp (overlay)
+  "Is 'overlay' an annotation?"
   (annotate-overlay-filled-p overlay))
 
 (defun annotate-annotate ()
--- a/annotate.el
+++ b/annotate.el
@@ -710,7 +710,8 @@ an overlay and it's annotation."
   (format "-%i,%i +%i,%i" start-line diff-size start-line diff-size)))
 
 (defun annotate-checksum-from-dump (record)
-  (nth 2 record))
+  (and (> (length record) 2)
+       (nth 2 record)))
 
 (defun annotate-annotations-from-dump (record)
   (nth 1 record))
--- a/annotate.el
+++ b/annotate.el
@@ -116,20 +116,20 @@
   :group 'annotate)
 
 (defcustom annotate-maximum-size-checksum 5000000
-  "Calculate checksum of  the current buffer only if  the size is
+  "Calculate checksum of the current buffer only if the size is
 less than this size (in characters)"
-  :type  'integer
+  :type 'integer
   :group 'annotate)
 
 (defconst annotate-warn-file-changed-control-string
   (concat "The file '%s' has changed on disk "
           "from the last time the annotations were saved.\n"
           "Chances are that they will not be displayed correctly")
-  "The message to  warn the user that file has  been modified and
+  "The message to warn the user that file has been modified and
   annotations positions could be outdated")
 
 (defcustom annotate-blacklist-major-mode '(org-mode)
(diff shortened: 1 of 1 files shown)


annotate.el
```diff
@@ -115,6 +115,43 @@
...
+less than this size (in characters)"
+  :type  'integer
+  :group 'annotate)
+
+(defconst annotate-warn-file-changed-control-string
+  (concat "The file '%s' has changed on disk "
+          "from the last time the annotations was saved. "
+          "Chances are that they will not be displayed correctly")
+  "The  message to  warn user  that  file has  been modified  and
+  annototatnions position could be outdated")
```
bastibe: ...the annotations *were* saved...
...to warn *the* user that *the* file has been modified and *annotation positions* could...

annotate.el
```diff
@@ -115,6 +115,43 @@
   :type 'string
   :group 'annotate)
 
+(defcustom annotate-maximum-size-checksum 5000000
+  "Calculate checksum of  the current buffer only if  the size is
+less than this size (in characters)"
+  :type  'integer
+  :group 'annotate)
```
bastibe: Why did you add this constraint?

annotate.el
```diff
@@ -115,6 +115,43 @@
...
+          "from the last time the annotations was saved. "
+          "Chances are that they will not be displayed correctly")
+  "The  message to  warn user  that  file has  been modified  and
+  annototatnions position could be outdated")
+
+(defcustom annotate-blacklist-major-mode '(org-mode)
+  "Prevent  loading of  annotated-mode  When  the visited  file's
+major mode is a member of this list (space separated entries)."
+  :type  '(repeat symbol)
+  :group 'annotate)
```
bastibe, line 129: This was discussed in a different issue as well. Do you know if the other issue made any progress on this?

annotate.el
```diff
@@ -115,6 +115,43 @@
...
+  (let ((annotate-allowed-p (with-current-buffer (current-buffer)
+                              (not (cl-member major-mode annotate-blacklist-major-mode)))))
+    (cond
+     ((not annotate-allowed-p)
+      (annotate-shutdown)
+      (setq annotate-mode nil))
+    (annotate-mode
+     (annotate-initialize))
+    (t
+     (annotate-shutdown)))))
```
bastibe, line 143: This function should not be called `-hook`, as that is commonly used for variables that may *contain* functions, not the functions themselves. How about `annotate-initialize-maybe`?

annotate.el
```diff
@@ -210,18 +255,26 @@
         (message "Annotations saved."))))
 
 (defun annotate-actual-comment-start ()
+  "String for  comment start  related to current  buffer's major
+mode."
```
bastibe, line 267: Thank you for adding documentation!

annotate.el
```diff
@@ -115,6 +115,43 @@
...
+(defconst annotate-warn-file-changed-control-string
+  (concat "The file '%s' has changed on disk "
+          "from the last time the annotations was saved. "
+          "Chances are that they will not be displayed correctly")
+  "The  message to  warn user  that  file has  been modified  and
+  annototatnions position could be outdated")
+
+(defcustom annotate-blacklist-major-mode '(org-mode)
+  "Prevent  loading of  annotated-mode  When  the visited  file's
+major mode is a member of this list (space separated entries)."
```
bastibe, line 127: ...loading of *annotate-mode*...

annotate.el
```diff
@@ -115,6 +115,43 @@
...
+    (annotate-mode
+     (annotate-initialize))
+    (t
+     (annotate-shutdown)))))
+
+(cl-defun annotate-buffer-checksum (&optional (object (current-buffer)))
+  (if (< (buffer-size)
+         annotate-maximum-size-checksum)
+      (md5 object)
+    nil))
```
bastibe: Does annotate.el already require `cl`? If not, could this be implemented without `cl`?

bastibe: Please add a docstring.

annotate.el
```diff
@@ -115,6 +115,43 @@
...
+
 (defun annotate-initialize ()
```
bastibe, line 157: Please add a docstring.

annotate.el
```diff
@@ -133,6 +170,11 @@
    nil '((annotate--font-lock-matcher (2 (annotate--annotation-builder))
                                       (1 (annotate--change-guard))))))
 
+(defun annotate-overlay-filled-p (overlay)
```
bastibe, line 177: Please add a docstring.

annotate.el
```diff
@@ -210,18 +255,26 @@
...
 
 (defun annotate-comments-length ()
+  "Total length of the comment markers (start and end) strings."
   (+ (string-width (annotate-actual-comment-start))
      (string-width (annotate-actual-comment-end))))
 
 (defun annotate-wrap-in-comment (&rest strings)
+  "Put comment  markers at  the start and  (if does  makes sense)
+ends  of   a  string.   See:   annotate-actual-comment-start  and
+annotate-actual-comment-end"
```
bastibe, line 285: ...(if *it* makes sense) *end*...

annotate.el
```diff
@@ -656,15 +709,32 @@ an overlay and it's annotation."
                       (1+ (- (line-number-at-pos end) (line-number-at-pos start))))))
   (format "-%i,%i +%i,%i" start-line diff-size start-line diff-size)))
 
+(defun annotate-checksum-from-dump (record)
```
bastibe, line 722: Please add docstring.

annotate.el
```diff
@@ -656,15 +709,32 @@ an overlay and it's annotation."
...
+  (and (> (length record) 2)
+       (nth 2 record)))
+
+(defun annotate-annotations-from-dump (record)
```
bastibe, line 728: Please add docstring.

bastibe (changes requested):
Thank you so much for this pull request! This is exactly what I had in mind in #6.

Before merging, I would request a few small changes. In addition to the changes mentioned in the review comments, please go through all changed docstrings and delete any double-spaces.

Again, thank you very much. I see you implemented it without breaking backwards compatibility, too! Great job!

cage2:
Hi Bastibe!

> Thank you so much for this pull request! This is exactly what I had in mind in #6.

Oh, that's great! :)
 
> Before merging, I would request a few small changes. In addition to the changes mentioned in the review comments, please go through all changed docstrings and delete any double-spaces.

Sure! No problem! :)

> 
> Again, thank you very much. I see you implemented it without breaking backwards compatibility, too! Great job!

To be honest i /tried/ not breaking the compatibility but i sadly discovered i failed to accomplish that. :(
Anyway i just pushed a quick hack that should fix that! Not the most elegant code i have seen but better that keeping a bug, i guess ;-)

Bye! 
C.

annotate.el
cage2: Thank you for ~~fix~~ fixing my bad english!

annotate.el
cage2: Because, i think, if the file is very large the calculation could take a lot of time so maybe an user could just skips this step and save time even if they know that annotation could be "out of sync".

annotate.el
cage2: I still have problem with global mode but i think adding this variable was a good idea anyway because the incompatibility with org-mode is a known problem and this setting could prevent a bit of frustration for an user that do not read documentation with attention (like me for example)! ;-D

annotate.el
cage2: nice!

annotate.el
cage2: i hate writing documentation (like many others :-)) but i hate more code that is not documented!  ;-D

annotate.el
cage2: IIRC cl-defun is needed because of the default optional argument

https://www.gnu.org/software/emacs/manual/html_mono/cl.html#Argument-Lists

So maybe the right question is: "is an optional argument needed at all?" :)

annotate.el
bastibe: I would suspect that md5 will never take remotely as long as Emacs will take to load the file in the first place. After all, md5 is routinely used to checksum gigabytes worth of data.

Did you have an actual example where big files were a problem? Otherwise, I'd say we go without the added complexity of this constraint (for now). We can still add them later if it turns out to be a problem for users.

If that's ok with you, of course.

annotate.el
bastibe: OK, cool.

annotate.el
bastibe: Silly me. I didn't even realize that annotate.el *already* requires `cl`. So this line of inquiry was pointless. Sorry about that.

bastibe:
As far as I can tell, the only open issue remains whether to remove the size restriction or not. I'll leave it up to you to keep it in the code or not.

Let me know when you think this is ready to merge!

annotate.el
cage2: Hello!

i am fine with removing the constraint, i have made a (very naive) test and -for a 5 years old computer- a ~200Mb file takes about 0.4s to calculate the hash and a couple of second (very approximate estimation!) to open in emacs, i am OK to remove the size check as i think is unusual to have more than a couple of files so big and i bet modern system got better calculating hash.

Maybe we could inform (in the README, for example) the user that such calculation occurs because, for example, i use desktop mode that restores all the buffers opened in a previous emacs session and is not unlikely (at least for me :)) to get restored more than 20 buffer or so. Anyway i just pushed the fix with the check removed and we can save the README update for another PR, if needed.

Bye!
C.

bastibe: Alright!

bastibe:
Thank you very much for your contribution, and our enjoyable conversation.

cage2:
My pleasure! :)

Bye!
C.

PR #44, opened 2019-09-28
- added an annotation summary window

cage2:
the  command   'annotate-show-annotation-summary'  popup   a  window
  containing a  textual description  of the  annotations found  in the
  annotation file (like ~/.emacs/annotations) with the format:

      -------

  /path/of/an/annotated/file

  - annotation text
  - annotation text

  /another/path/of/an/annotated/file

  - annotation text

  ...

      -------

Each annotation text is a link  that, if activated (via a mouse click,
for  example),  opens a  buffer  showing  the  annotated file  at  the
position  of  the  annotated  text (the  underlined  one)  linked  to
annotation.

Bye!!
C.

### DIFF
--- a/annotate.el
+++ b/annotate.el
@@ -808,7 +808,7 @@ essentially what you get from:
            (modified-p           (buffer-modified-p)))
       (if (old-format-p annotation-dump)
           (annotate-load-annotation-old-format)
-        (when (and (not (old-format-p annotations))
+        (when (and (not (old-format-p annotation-dump))
                    old-checksum
                    new-checksum
                    (not (string= old-checksum new-checksum)))
--- a/annotate.el
+++ b/annotate.el
@@ -117,6 +117,12 @@
   :type 'string
   :group 'annotate)
 
+(defcustom annotate-blacklist-major-mode '(org-mode)
+  "Prevent loading of annotate-mode When the visited file's
+major mode is a member of this list (space separated entries)."
+  :type  '(repeat symbol)
+  :group 'annotate)
+
 (defconst annotate-warn-file-changed-control-string
   (concat "The file '%s' has changed on disk "
           "from the last time the annotations were saved.\n"
@@ -124,14 +130,14 @@
   "The message to warn the user that file has been modified and
   annotations positions could be outdated")
 
-(defcustom annotate-blacklist-major-mode '(org-mode)
-  "Prevent loading of annotate-mode When the visited file's
-major mode is a member of this list (space separated entries)."
-  :type  '(repeat symbol)
-  :group 'annotate)
-
-(defconst annotate-summary-list-prefix "- "
+(defconst annotate-summary-list-prefix "  - "
   "The string used as prefix for each text annotation item in summary window")
+
+(defconst annotate-summary-list-prefix-file "* File: "
+  "The string used as prefix for each annotated file item in summary window")
+
+(defconst annotate-summary-list-prefix-snippet "** Annotated text: "
+  "The string used as prefix for each annotation snippet item in summary window")
 
 (defconst annotate-ellipse-text-marker "..."
   "The string used when a string is truncated with an ellipse")
@@ -750,6 +756,13 @@ essentially what you get from:
 (annotate-annotations-from-dump (annotate-load-annotations))). "
   (cl-first annotation))
 
+(defun annotate-ending-of-annotation (annotation)
+  "Get the ending point of an annotation. The arg 'annotation' must be a single
+annotation field got from a file dump of all annotated buffers,
+essentially what you get from:
+(annotate-annotations-from-dump (annotate-load-annotations))). "
+  (cl-second annotation))
+
 (defun annotate-text-of-annotation (annotation)
   "Get the text of an annotation. The arg 'annotation' must be a single
 annotation field got from a file dump of all annotated buffers,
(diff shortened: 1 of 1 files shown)


annotate.el
```diff
@@ -896,5 +922,48 @@ file."
...
+(defun annotate-summary-button-pressed (button)
+  "Callback called when an annotate-summary-button is activated"
+  (let ((buffer (find-file-other-window (button-get button 'file))))
+    (with-current-buffer buffer
+      (goto-char (button-get button 'go-to)))))
+
+(defun annotate-show-annotation-summary ()
+  "Show a summary of all the annotations in a temp buffer"
+  (interactive)
+  (cl-labels ((ellipsize (text)
```
bastibe: Why is this an inline function instead of a `defun`?

annotate.el
```diff
@@ -896,5 +922,48 @@ file."
...
+                         (if (> (string-width text)
+                                annotate-summary-link-max-width)
+                             (concat (subseq text 0 (- annotate-summary-link-max-width 3))
+                                     "...")
+                           text)))
+    (with-temp-buffer-window
+     "*annotations*" nil nil
+     (with-current-buffer "*annotations*"
```
bastibe: Are both `with-temp-buffer-window` and `with-current-buffer` necessary, or would one of them suffice?

annotate.el
```diff
@@ -730,6 +737,25 @@ file."
 file."
   (nth 1 record))
 
+(defun annotate-filename-from-dump (record)
+  "Get the filename field from an annotation list loaded from a
+file."
+  (cl-first record))
+
+(defun annotate-start-annotation-dump (annotation)
```
bastibe: "start" sounds like a verb. Similarly to the previous function, I would propose something like `annotate-beginning-of-annotation`.

annotate.el
```diff
@@ -730,6 +737,25 @@ file."
...
+  "Get the starting point of an annotation. The arg 'annotation' must be a single
+annotation field got from a file dump of all annotated buffers,
+essentially what you get from:
+(annotate-annotations-from-dump (annotate-load-annotations))). "
+  (cl-first annotation))
+
+(defun annotate-text-annotation-dump (annotation)
```
bastibe: See above: How about `annotate-text-of-annotation`?

annotate.el
```diff
@@ -128,6 +130,11 @@ major mode is a member of this list (space separated entries)."
   :type  '(repeat symbol)
   :group 'annotate)
 
+(defcustom annotate-summary-link-max-width  64
```
bastibe: Why is it 64, instead of, say 72, 80, or `fill-column`, or `visual-wrap-column` (if set)?

bastibe (changes requested):
That's really cool! I love it!

Could you post a screenshot of what it looks like, when rendered in your Emacs?

What are your thoughts on including an entire (truncated) source line instead of just the annotated text?

I have included a few comments on how to improve the code a bit prior to merging. 

Again, thank you very much!

annotate.el

Well, i just thought this function could not be useful outside `annotate-show-annotation-summary`, if you prefer i can just refactor it into `annotate-ellipsize` or something like that.

annotate.el
cage2: I think `with-temp-buffer-window` is needed so I tried to remove `with-current-buffer` and the summary text was inserted in the selected buffer instead of the summary window just created (i mean, the one that should hosts the summary), in my case the text was inserted straight in the source code, that was creepy! :)

annotate.el
cage2: OK!

annotate.el
cage2: no problem!

annotate.el
cage2: Because it is a power of 2! ;-)

Just kidding, but yes, it was totally arbitrary. I wonder if i could just use `window-body-width` and getting rid of this variable instead.

cage2: Seems that `window-body-width` can be used! Is this solution ok for you?

cage2:
> That's really cool! I love it!
> 
> Could you post a screenshot of what it looks like, when rendered in your Emacs?

Sure, no problem!

> What are your thoughts on including an entire (truncated) source line instead of just the annotated text?

yes, this would be a nice improvements, any idea of the layout to accomodate this text?

Something like that, for example?

``` text

/path/of/an/annotated/file

 *  annotated text snippet 
    - annotation text
    - annotation text
 *  another annotated text snippet 
    -   annotation text

```

and so on

> Again, thank you very much!

You're welcome! I feel i am starting to learning a little more about emacs and elisp!

Bye!
C.

annotate.el
bastibe: OK, interesting.

annotate.el
bastibe: I think I'd prefer that, yes. I understand your intention, but factoring out the function would improve readability, I think.

annotate.el
bastibe: Absolutely! Thank you for investigating that!

Just out of curiosity, does the summary work correctly with crazy unicode shenanigans, like emoji or combining diacritics? It doesn't have to, if you ask me, but these kinds of things always were problematic for annotate.el.

bastibe:
> > What are your thoughts on including an entire (truncated) source line instead of just the annotated text?
> 
> yes, this would be a nice improvements, any idea of the layout to accommodate this text?

Good question. Possibly something like

```
/path/to/file
* this is a line containing an {annotated text snippet} and some more...
  - annotation text
```

We could highlight the annotated text either with ascii characters, or with overlays, just like annotate.el itself does it.

Come to think of it, the coolest solution would be an `occur-edit-mode` style buffer that lists the original, editable, source lines with their true annotations. But frankly, I have no idea whatsoever how to accomplish that, or if it is even possible.

annotate.el
cage2: well, seems that with a couple of expression more i was able to get rid of the second macro (`with-current-buffer`)!

cage2:
> > > What are your thoughts on including an entire (truncated) source line instead of just the annotated text?
> > 
> > 
> > yes, this would be a nice improvements, any idea of the layout to accommodate this text?
> 
> Good question. Possibly something like
> 
> ```
> /path/to/file
> * this is a line containing an {annotated text snippet} and some more...
>   - annotation text
> ```
> We could highlight the annotated text either with ascii characters, or with overlays, just like annotate.el itself does it.

I discovered (probably re-discovered) `outline-mode` and this is what i got so far:

https://www.autistici.org/interzona/img/misc/annotate-summary-example.png

I think this is close to what you asked, i have added some more empty lines here and there just according to my very personal task, but you could have different opinion about that of course! :)
 
> Come to think of it, the coolest solution would be an `occur-edit-mode` style buffer that lists the original, editable, source lines with their true annotations. But frankly, I have no idea whatsoever how to accomplish that, or if it is even possible.

This is a very ambitious goal for sure, i do not even know where to start to accomplish that, to be honest! :)

Bye!

annotate.el
cage2: As you can see i have added some more local function instead! :)
My opinion is: as they are very specific to the enclosing function the use of `cl-label` allows me to write code without "pollute" the global namespace and keeping the name reasonably short (without the `annotation-` prefix), i would say this is exactly the reason the local function form exists! But using or not local functions is a matter of personal taste so: i suggest to keep the local functions here but i left this decision to you.

Bye!
C.

annotate.el
cage2: I would say is not a mess but there are rooms for improvements! :)

annotate.el
bastibe: I'm OK with your current solution. It does make the code somewhat cleaner.

annotate.el
```diff
@@ -896,5 +936,80 @@ file."
...
+               (filename        (annotate-filename-from-dump annotation)))
+           (when (not (null all-annotations))
+             (insert (format (concat annotate-summary-list-prefix-file "%s\n\n")
+                             filename))
+             (dolist (annotation-field all-annotations)
+               (let* ((button-text      (format "%s"
+                                                (annotate-text-of-annotation annotation-field)))
+                      (annotation-begin (annotate-beginning-of-annotation annotation-field))
+                      (annotation-end   (annotate-ending-of-annotation    annotation-field))
+                      (snippet-text     (build-snippet)))
```
bastibe: In general, I prefer functions to have explicit arguments instead of referring to local dynamic bindings. Is there a specific reason why you implemented `build-snippet` this way?

bastibe:
This looks really cool!

Why are the annotation texts list items? Do you commonly use multiple annotations of the same text? I would have expected each text to have exactly one annotation, which would not require a list.

cage2:
> This looks really cool!

Thank you! :)
 
> Why are the annotation texts list items? Do you commonly use multiple annotations of the same text? I would have expected each text to have exactly one annotation, which would not require a list.

Sorry, the list bullet was just wrong, left there only because i forgot to remove it. Now I changed it with just  spaces.

Bye!
C.

annotate.el
cage2: Yes, this is confusing at best, fixed!

bastibe:
Thank you very very much for all your work on this!

The pull request looks good to me, now! Would you like me to merge it?

cage2:
On Tue, Oct 08, 2019 at 12:40:32AM -0700, Bastian Bechtold wrote:

Hi Bastibe!

> Thank you very very much for all your work on this!

My pleasure!

> The pull request looks good to me, now! Would you like me to merge it?

I you are positive: sure!

Moreover i have (almost) ready a  couple of pull requests unrelated to
this one but i think i need your help to understand if they are a good
idea or not! :)

Bye!
C.

cage2:
Is there time for a last minute fix? ;-)

bastibe:
I'll merge when you tell me you're ready. There's no rush.

> Moreover i have (almost) ready a  couple of pull requests unrelated to this one but i think i need your help to understand if they are a good idea or not! :)

Just open a pull request and I'll have a look!

Thank you for putting so much work into this project!

cage2:
On Wed, Oct 09, 2019 at 01:21:57AM -0700, Bastian Bechtold wrote:

Hi!

> I'll merge when you tell me you're ready. There's no rush.

Ok, i am fine with merging after the last commit! :)
I think we could even made this mode global now.

Bye!!
C.

cage2:
On Thu, Oct 10, 2019 at 12:51:36AM -0700, Bastian Bechtold wrote:
> Merged #44 into master.

Thank you!

C.

PR #45, opened 2019-10-10
Preventing loss of undo list

cage2:
According  to the  documentation  the function  'uffer-disable-undo'
  discards all  the already registered buffer's  modification. This is
  undesiderable, to  prevent this  the old buffer  list is  copied and
  restored in 'annotate--remove-annotation-property'.


bastibe:
Cool! Thank you very much!

PR #47, opened 2019-10-22
Some other changes

cage2:
Same more changes: bugfix and a little new feature! :)
Bye!
C.

### DIFF
--- a/annotate.el
+++ b/annotate.el
@@ -881,14 +881,14 @@ essentially what you get from:
 (annotate-annotations-from-dump (annotate-load-annotations))). "
   (cl-second annotation))
 
-(defun annotate-text-of-annotation (annotation)
+(defun annotate-annotation-string (annotation)
   "Get the text of an annotation. The arg 'annotation' must be a single
 annotation field got from a file dump of all annotated buffers,
 essentially what you get from:
 (annotate-annotations-from-dump (annotate-load-annotations))). "
   (nth 2 annotation))
 
-(defun annotate-sample-text-of-annotation (annotation)
+(defun annotate-annotated-text (annotation)
   "Get the annotated text of an annotation. The arg 'annotation' must be a single
 annotation field got from a file dump of all annotated buffers,
 essentially what you get from:
@@ -912,9 +912,9 @@ essentially what you get from:
     (when (not (null annotations))
       (save-excursion
         (dolist (annotation annotations)
-          (let ((start (annotate-beginning-of-annotation annotation))
-                (end   (annotate-ending-of-annotation    annotation))
-                (text  (annotate-text-of-annotation      annotation)))
+          (let ((start              (annotate-beginning-of-annotation annotation))
+                (end                (annotate-ending-of-annotation    annotation))
+                (annotation-string  (annotate-annotation-string       annotation)))
             (annotate-create-annotation start end text)))))
     (set-buffer-modified-p modified-p)
     (font-lock-fontify-buffer)
@@ -950,11 +950,11 @@ essentially what you get from:
         (annotations
          (save-excursion
            (dolist (annotation annotations)
-             (let ((start  (annotate-beginning-of-annotation   annotation))
-                   (end    (annotate-ending-of-annotation      annotation))
-                   (text   (annotate-text-of-annotation        annotation))
-                   (sample (annotate-sample-text-of-annotation annotation)))
-               (annotate-create-annotation start end text sample))))))
+             (let ((start  (annotate-beginning-of-annotation annotation))
+                   (end    (annotate-ending-of-annotation    annotation))
+                   (text   (annotate-annotation-string       annotation))
+                   (note   (annotate-annotated-text          annotation)))
+               (annotate-create-annotation start end text note))))))
         (set-buffer-modified-p modified-p)
         (font-lock-fontify-buffer)
         (when annotate-use-messages
@@ -982,19 +982,25 @@ essentially what you get from:
   (or (null a)
       (string= "" a)))
 
-(defun annotate-create-annotation (start end &optional text sample)
- "Create a new annotation for selected region. If this function
- is called from procedure 'annotate-load-annotations' the
- argument 'sample' should be not null. In this case we know that
- an annotation existed in a text interval defined in the database
- metadata (the database located in the file specified by the
- variable 'annotate-file') and should just be restored. Sometimes
- the annotated text can not be found in said interval because the
- annotated file's content changed and annotate-mode could not
- track the changes (e.g. save the file when annotate-mode was not
- active/loaded) in this case the matching text is searched in a
- region surrounding the interval and, if found, the buffer is
- annotated right there."
+(defun annotate-create-annotation (start end &optional annotation-text annotated-text)
+  "Create a new annotation for selected region.
+
+  Here the argument 'annotation-text' is the string that appears in the margin of the window
+  and 'annotated-text' is the string that is underlined.
+
+  If this function is called from procedure
+  'annotate-load-annotations' the argument 'annotated-text'
+  should be not null. In this case we know that an annotation
+  existed in a text interval defined in the database
+  metadata (the database located in the file specified by the
+  variable 'annotate-file') and should just be
+  restored. Sometimes the annotated text (see above) can not be
+  found in said interval because the annotated file's content
+  changed and annotate-mode could not track the
+  changes (e.g. save the file when annotate-mode was not
+  active/loaded) in this case the matching
+  text ('annotated-text') is searched in a region surrounding the
+  interval and, if found, the buffer is annotated right there."
   (cl-labels ((create-annotation     (start end annotation-text)
                                      (let ((highlight (make-overlay start end)))
                                        (overlay-put highlight 'face 'annotate-highlight)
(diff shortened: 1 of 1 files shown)


annotate.el
```diff
@@ -871,6 +879,14 @@ essentially what you get from:
 (annotate-annotations-from-dump (annotate-load-annotations))). "
   (nth 2 annotation))
 
+(defun annotate-sample-text-of-annotation (annotation)
```
bastibe: These function names are confusing. I think we should find a better naming scheme for `text-of-annotation` and `sample-text-of-annotation`. 

Maybe `annotate-destination-text-of-annotation` and `annotate-annotation-text-of-annotation`?

It's not perfect yet...

annotate.el
```diff
@@ -950,18 +968,65 @@ essentially what you get from:
...
-      (when (use-region-p)
-        (deactivate-mark))))
-  (save-excursion
-    (goto-char end)
-    (font-lock-fontify-block 1)))
+  (cl-labels ((create-annotation (start end annotation-text)
+                                 (let ((highlight (make-overlay start end)))
+                                   (overlay-put highlight 'face 'annotate-highlight)
+                                   (overlay-put highlight 'annotation annotation)))
+              (move-lines        (start line-count)
```
bastibe: How about `beginning-of-nth-line`?

annotate.el
```diff
@@ -950,18 +968,65 @@ essentially what you get from:
...
+                                 (save-excursion
+                                   (goto-char start)
+                                   (forward-line line-count)
+                                   (beginning-of-line)
+                                   (point)))
+              (go-backward       (start)
```
bastibe: `beginning-of-previous-line`? But why two lines?

annotate.el
```diff
@@ -950,18 +968,65 @@ essentially what you get from:
...
+                                 (move-lines start -2))
+              (go-forward        (start)
```
bastibe: `beginning-of-next-line`? But why two lines?

annotate.el
bastibe: Oh, I see. It's two lines because that's the "search region" where a moved annotation can be, right? It might be useful to make that a customizable variable.

bastibe (changes requested):
Cool! Finally, we'll have some kind of plausibility check that the annotations are still valid.

(I'd have appreciated a more descriptive pull request text ;-) )

annotate.el
```diff
@@ -950,18 +968,65 @@ essentially what you get from:
...
     (set-buffer-modified-p modified-p)))
 
-(defun annotate-create-annotation (start end &optional text)
+(defun annotate-string-empty-p (a)
+  "Is the arg an empty string or null?"
+  (or (null a)
+      (string= "" a)))
+
+(defun annotate-create-annotation (start end &optional text sample)
   "Create a new annotation for selected region."
```
bastibe: Docstring probably needs to be expanded with new functionality.

annotate.el
cage2: sure! Changed!

annotate.el
cage2: Good idea!

cage2: > (I'd have appreciated a more descriptive pull request text ;-) )

Sorry!

annotate.el
cage2: So if i understand correctly:

``` text
 from 'annotate-sample-text-of-annotation' to  'annotate-annotation-text-of-annotation'
 from 'annotate-text-of-annotation'        to  'annotate-destination-text-of-annotation'
```
right?

annotate.el
cage2: I tried, not very happy with the results, though! :(

annotate.el
bastibe: Yeah, but now that you write them out, those are not good names yet.

So we have to differentiate between the text being annotated, and the annotation text. Maybe that's the key right there: "annotation text" vs. "annotated text". And maybe mix up that "text" to make them more distinct:

    from annotate-sample-text-of-annotation   to annotate-annotated-text
    from annotate-text-of-annotation          to annotate-annotation-string

(And possibly reword other occurrences accordingly)

What do you think about that?

annotate.el
bastibe: Iif I'm reading the code correctly, it does

- If no text or sample are given, create a new annotation between start and end, and ask for the annotation text
- If a text is given but no sample, create a new annotation between start and end
- If both text and sample are given, and the sample matches the text between start and end, create a new annotation between start and end
- If both text and sample are given, and the sample does not match the text between start and end, search the surrounding area for a matching location, and create a new annotation there
- If both text and sample are given, and the sample does not match the text between start and end, and no matching location is nearby, silently ignore the annotation

It might be useful to split this into multiple functions (as a general rule of thumb, I try to split stuff into multiple functions if I start to feel compelled to add an "and" to the function name).

For example, we could make `text` and `sample` non-optional by changing `annotate-annotate` to grab them before calling `annotate-create-annotation`.

And then maybe add a failure condition for non-matching annotations, probably as a warning.

annotate.el
cage2: Hi bastibe, and welcome back! :-)

I agree with the changes and i am going to push changes accordingly to your suggestions. I have to admit i am a bit confused between annotated text and annotation text i hope i did right with this changes. :-D

Bye!
 C.

annotate.el
Yes i think you understand correctly the goal of the function (and i hope my code works this way :-)) 

Unfortunately i think i do not understand very well how to refactor the function, instead. :)
The two optional arguments in ``annotate-create-annotation`` are used only when the annotations are restored from the database i can not see how they can be used in `annotate-annotate`, should we change: `(annotate-create-annotation start end)` to `(annotate-create-annotation start end nil nil)`? Please help me to understand.

Anyway, good idea for the warning if annotation matching fails! I like it! :)

Bye!

bastibe: Hi!

When looking through the code, I noticed that all but one place calls `annotate-create-annotation` with four arguments. If that one place, `annotate-annotate`, would use all four arguments, we could make the last two arguments non-optional. So I propose to change `annotate-annotate` to supply the last two arguments, and to make them non-optional. This will clean up the implementation of `annotate-create-annotation` somewhat, because we won't have to include all the checks for the last two arguments any longer.

Does that explain it better?

cage2: Hello!!

yes i think the fog in my brain is starting to going away! :-)

so, i just made the two arguments non optional and now i should move all that stuff:

``` lisp 
   (let ((annotation (or annotation-text
                          (read-from-minibuffer "Annotation: "))))
      (when (not (or (null annotation) 
                     (string= "" annotation))) ; note this can be changed with
                                               ; (not (string-empty-p annotation))
        (if (not (annotate-string-empty-p annotated-text))
 ```
up to `annotate-annotate`, is it?

Bye!
C.

bastibe: That's the idea!

Except of course you won't need all the conditionals, as `annotate-annotate` knows it has to ask for the annotation-string and can grab the annotated-text from the buffer.

cage2:
On Thu, Nov 07, 2019 at 11:19:21AM -0800, Bastian Bechtold wrote:

Hello!


>
> That's the idea!
> Except  of  course   you  won't  need  all   the  conditionals,  as
> `annotate-annotate` knows  it has  to ask for  the annotation-string
> and can grab the annotated-text from the buffer.

Very well!

I think (an hope) this commit:

https://github.com/bastibe/annotate.el/pull/47/commits/3108c44c5308f6d26e3cb143bf38cc69e300298a

should implement your suggestion.

To me this change does make a lot of sense!

Bye!
C.

bastibe:
Absolutely! Great work!

bastibe:
As far as I can tell, this is almost complete! I'd just add an error message if the annotation couldn't be loaded and then this looks good to merge!

cage2:
On Fri, Nov 08, 2019 at 06:46:32AM -0800, Bastian Bechtold wrote:
> Absolutely! Great work!

Thank you very much! :)

Bye
C.

cage2:
On Fri, Nov 08, 2019 at 06:49:59AM -0800, Bastian Bechtold wrote:

Hi!

> As far as I can tell, this is almost complete! I'd just add an error
> message if the annotation couldn't be loaded

Sure, i am going to add it, seems to me it should not be a complicated
goal to accomplish so i hope to add int in the next hour.

> and then this looks good to merge!

Also  i would  like  to prevent  to  show a  summary  window when  the
metadata db is  empty or contain no actual annotations,  when the former
happens (empty window) this looks a bit not elegant to me.  In this case
i would prefer just to print a message to inform the user that there are no
annotations to show.  Are you OK with adding this feature?

Bye!!

C.

cage2:
Well, seems i have pushed the changes for the checks about the summery window, sorry! 
Anyway if you do not like it just tell me and i will revert the last commit.

Bye!!!
C.

bastibe (approved):

bastibe:
Cool, I like it!

Thank you very much for another great contribution.

Feel free to merge when you're ready!

cage2:
On Sat, Nov 09, 2019 at 04:44:59AM -0800, Bastian Bechtold wrote:
> Cool, I like it!
>
> Thank you very much for another great contribution.
>
> Feel free to merge when you're ready!

With fixing this macro i think we can merge the two branches.

Bye!
C.

cage2:
An that's all, for now! :D

bastibe:
🎉

Issue #48, opened 2019-10-30
[feature request] enable annotating Info manual


randomwangran:
Dear the maintainers of the package of `annotate.el`:

Thanks for writing this package, I find it useful when reading
computer programs. It makes my ideas seamlessly intergrated with
the sources very well.

My initial idea was to use this package to annotate the Info
manual. I found that it is very enjoyable to read the Elisp
manual inside the Info mode, so I thougth it would be better if I
could annotate my thoughts directly into the Info manual.

I tweaked some of the functions before Sept. 24. I was able to
achieve the function. Please see the following GIFs. I did not
report to you because I thought my code was not robust.

1.  Able to write notes in Info mode:
![a](https://user-images.githubusercontent.com/8197831/67827522-a4541600-faa6-11e9-9e45-0ff4ab2a2cb6.gif)
2. Able to write notes in the sub-nodes:
![b](https://user-images.githubusercontent.com/8197831/67827526-a6b67000-faa6-11e9-9689-7e1952a029a6.gif)
3. After reopen Emacs, it can reload all the notes in the Info manual:
![c](https://user-images.githubusercontent.com/8197831/67827528-a8803380-faa6-11e9-8eab-c2ed1615f86d.gif)

There are many commits after Sept. 24 and after that my tweaks
were not compatable with your code.

I was wondering that if someone could make a decent pull request
to this package.

Followings are my tweaks, and I hope it can help someone
interested in making it.

Thanks for your effort. Really appreciate it.

```elisp
(add-to-list 'load-path "~/.emacs.d/annotate.el")
(require 'annotate)
(setq annotate-file (expand-file-name "~/Dropbox/myNote/annotations"))


;; for info file

;;;###autoload
(define-minor-mode annotate-mode
  "Toggle Annotate mode."
  :init-value nil
  :lighter " Ann"
  :keymap (make-sparse-keymap)
  :group 'annotate
  :after-hook (if annotate-mode
                  (progn (annotate-initialize) (split-window-below) (delete-other-windows))
                (annotate-shutdown)))

(defun tweak-annotate-load-annotations ()
  "Load all annotations from disk."
  (interactive)
  (let ((annotations (cdr (assoc-string
                           (substring-no-properties (or Info-current-file ""))
                           (annotate-load-annotation-data))))
        (modified-p (buffer-modified-p)))
    ;; remove empty annotations created by earlier bug:
    (setq annotations (cl-remove-if (lambda (ann) (eq (nth 2 ann) nil))
[...]

PR #49, opened 2019-11-12
- added features: annotate info documents.

cage2:
Hello!

this is my attempt to address #48 
there are some points to investigate in my opinion, in details:

- when an info node is annotated and the user try to save the file emacs will ask a file name where dumps the info buffer and this package annotate that saved file, if just 'annnotate-save-annotation' is called the correct file is annotated, though. i am not sure this is an actual bug, this behaviour seems reasonably correct to me; please @randomwangran, can you tell us it this is the same way you used your patched version of this program?

- the function 'annotate-info-select-fn' is a bit slow it is not difficult to understand why ;-);

- on my system if i do not add the test '(when (buffer-file-name)' in the hook called before a buffer's content is changed a weird error complaining about buffer encoding pop up, i suspect there is a better way to get rid of this error but i had no better idea so far.

Bye!
C.

### DIFF
--- a/annotate.el
+++ b/annotate.el
@@ -228,24 +228,23 @@ it is called any time the buffer content is changed (so, for
 example, text is added or deleted). In particular, it will
 rearrange the overlays bounds when an annotated text is
 modified (for example a newline is inserted)."
-  (when (buffer-file-name)
-    (annotate-with-inhibit-modification-hooks
-     (save-excursion
-       (let* ((bol (annotate-beginning-of-line-pos))
-              (eol (annotate-end-of-line-pos))
-              (ov  (cl-remove-if-not 'annotationp
-                                     (overlays-in bol eol))))
-         (dolist (overlay ov)
-           (annotate--remove-annotation-property (overlay-start overlay)
-                                                 (overlay-end   overlay))
-           ;; move the overlay if we are breaking it
-           (when (<= (overlay-start overlay)
-                     a
-                     (overlay-end overlay))
-             (move-overlay overlay (overlay-start overlay) a)
-             ;; delete overlay if there is no more annotated text
-             (when (annotate-annotated-text-empty-p overlay)
-               (delete-overlay overlay)))))))))
+  (annotate-with-inhibit-modification-hooks
+   (save-excursion
+     (let* ((bol (annotate-beginning-of-line-pos))
+            (eol (annotate-end-of-line-pos))
+            (ov  (cl-remove-if-not 'annotationp
+                                   (overlays-in bol eol))))
+       (dolist (overlay ov)
+         (annotate--remove-annotation-property (overlay-start overlay)
+                                               (overlay-end   overlay))
+         ;; move the overlay if we are breaking it
+         (when (<= (overlay-start overlay)
+                   a
+                   (overlay-end overlay))
+           (move-overlay overlay (overlay-start overlay) a)
+           ;; delete overlay if there is no more annotated text
+           (when (annotate-annotated-text-empty-p overlay)
+             (delete-overlay overlay))))))))
 
 (defun annotate-info-select-fn ()
   (annotate-clear-annotations)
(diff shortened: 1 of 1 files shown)


annotate.el
```diff
@@ -228,30 +228,37 @@ it is called any time the buffer content is changed (so, for
...
-                                               (overlay-end   overlay))
-         ;; move the overlay if we are breaking it
-         (when (<= (overlay-start overlay)
-                   a
-                   (overlay-end overlay))
-           (move-overlay overlay (overlay-start overlay) a)
-           ;; delete overlay if there is no more annotated text
-           (when (annotate-annotated-text-empty-p overlay)
-             (delete-overlay overlay))))))))
+  (when (buffer-file-name)
```
bastibe: We should investigate why this is necessary, as this effectively breaks annotate-mode if a file without a name is being edited.

Could you tell us the error message you are getting if this is left in?

cage2:
On Wed, Nov 13, 2019 at 12:00:00AM -0800, Bastian Bechtold wrote:

Hello Bastian!

[...]

> +  (when (buffer-file-name)
>

> We should  investigate why  this is  necessary, as  this effectively
> breaks annotate-mode if a file without a name is being edited.

Well      seems       that      i      found       the      "culprit":
`annotate--remove-annotation-property`.

What i think  i have understood so  far: when an info  node is loaded,
except when the root  node is loaded the first time  (i mean when `M-x
info` is called),  the *info* buffer is emptied and  filled again whit
the new  content's node.  This will  trigger `before-change-functions`
and, therefore, `annotate-before-change-fn`.   This function, in turn,
will call `annotate--remove-annotation-property` just after the buffer
has  been emptied.  The problem  is that,  for some  reason i  can not
understand,  the annotation  overlays are  still there  (in the  empty
buffer) from point 1  to point 1 (not a typo) :-/ in  a buffer of size
0...

so, unsurprisingly, this form fails: :-(

``` lisp
(remove-text-properties (point) (1+ (point)) '(display nil)))
```

So i have just removed the test for filename and added a test to check
that we are  not trying to remove  a string property in  a region that
does not exists because the buffer is empty.

I wonder why annotations exists at all in an empty buffer, though. :(

There  was  a problem  with  info  filename  that prevented  the  text
snippet, in  summary window, to shown  at all with an  annotated info
file. I fixed it and now the text appears but is all messed up...

There is still work to do i guess. :)

Bye!
C.

bastibe:
Fascinating! So, annotations have two components: Overlays (the red underlines), and a display property at the end of the line that shows the annotation. It sounds like the display property remains, even though the overlay is correctly deleted. How strange.

cage2:
On Thu, Nov 14, 2019 at 12:56:45AM -0800, Bastian Bechtold wrote:

> Fascinating! So, annotations have  two components: Overlays (the red
> underlines), and  a display  property at  the end  of the  line that
> shows the annotation.  It sounds like the  display property remains,
> even though the overlay is correctly deleted. How strange.

yes it is! :)

Turned  out that  the problem  was more  difficult than  expected, but
difficults are parts of the fun i guess! ;-D

Bye!
C.

bastibe:
As the complications of this pull request expand, I think it would be useful to clearly mark info-related code paths as such. Could you add comments and denote functions as `annotate-info-...` to make this clear?

cage2:
On Fri, Nov 15, 2019 at 12:17:50AM -0800, Bastian Bechtold wrote:

Hi Bastian!!

> As the complications  of this pull request expand, I  think it would
> be useful to clearly mark info-related code paths as such. Could you
> add  comments and  denote functions  as `annotate-info-...`  to make
> this clear?

Yes, of course!

Bye!
C.

cage2:
Hi bastibe!!

> 
> As the complications  of this pull request expand, I  think it would
> be useful to clearly mark info-related code paths as such. Could you
> add  comments and  denote functions  as `annotate-info-...`  to make
> this clear?

Is it OK now?

Bye!
C.

bastibe:
Cool, this looks good! merge when you're ready.

cage2:
Hi!

Very well, I just committed  a simple modification with 
https://github.com/bastibe/annotate.el/pull/49/commits/75d291a69fb4e10acbafdc7c7f11e4046cec71c7, so tomorrow i will merge that branch into master and close the issue #48.

Bye!!!
C.

cage2:
Ok, merged! Thank you for your help and support bastibe! :)

bastibe:
Thank you for all your hard work!

cage2:
On Wed, Nov 20, 2019 at 03:07:04AM -0800, Bastian Bechtold wrote:
> Thank you for all your hard work!

You're welcome! :)

Bye!
C.

randomwangran:
I was unable to access my productive computer last month, so I
could not respond to you.

I just did some tests and I think this feature has been
successfully implemented.

Thank you cage2 and bastibe.

cage2:
On Wed, Nov 27, 2019 at 06:52:01AM -0800, randomwangran wrote:

Hi!

> I was unable to access my productive computer last month, so I
> could not respond to you.

No problem!

> I just did some tests and I think this feature has been
> successfully implemented.
>
> Thank you cage2 and bastibe.

Happy to know that everything seems to works fine!

Enjoy this  package, and  thanks for your  suggestion to  improve this
software!

Bye!
C.

PR #50, opened 2019-11-23
- used a better regex when checking if a file is in info format.

cage2:
Hi!

According to the manual the separator for info format is a bit more complex than a simple `C-_`,  using the latter made the annotate file source looks like an info file! :D

Bye!
C.

### DIFF
--- a/annotate.el
+++ b/annotate.el
@@ -1243,7 +1243,8 @@ sophisticated way than plain text"
                               (t
                                (let* ((file-contents     (file-contents))
                                       (has-info-p        (string-match "info" filename))
-                                      (has-separator-p   (string-match "" file-contents))
+                                      (has-separator-p   (string-match "\\(\\)?\\(\\)?$"
+                                                                       file-contents))
                                       (has-node-p        (string-match "Node:" file-contents)))
                                  (if (or (annotate-info-root-dir-p filename)
                                          (and has-separator-p


bastibe:
So this matches "$" or "()$" or "()()$"? And before it matched the empty string? This all sounds wrong. I am not understanding something, obviously ;-)

Could you include a comment with a link to the info format or something like that?

cage2:
On Sun, Nov 24, 2019 at 11:52:19PM -0800, Bastian Bechtold wrote:

Hi Bastian!

> So this matches  "$" or "()$" or "()()$"? And  before it matched the
> empty  string?  This  all  sounds  wrong.  I  am  not  understanding
> something, obviously ;-)

I scratched my  head for a few minutes wondering  what happened to the
regexp  and  then i  realized  github  was  eating the  non  printable
characters! :D :D :D

https://raw.githubusercontent.com/bastibe/annotate.el/50ba1d719cadd9e2fe90ca2b536b57ade0b22826/annotate.el

>
> Could you include a comment with a link to the info format or something like that?

Sure!

Here is the source for the regex.

https://www.gnu.org/software/texinfo/manual/texinfo/html_node/Info-Format-Specification.html#Info-Format-Specification

I wonder if i made the right thing using "$" for the newline.

Bye!
C.

bastibe:
Oh! I see! that makes a lot more sense now.

Is there an escape sequence we can use instead of writing the unprintable characters into the file directly?

cage2:
On Tue, Nov 26, 2019 at 01:31:00AM -0800, Bastian Bechtold wrote:

Hi Bastian!

> Oh! I see! that makes a lot more sense now.

>  Is  there an  escape sequence  we can  use instead  of writing  the
> unprintable characters into the file directly?

This is a very good question, indeed!

Actually i do not know, i am going  to do a web search and report here
what i am going to find.


Bye!
C.

cage2:
On Tue, Nov 26, 2019 at 01:31:00AM -0800, Bastian Bechtold wrote:

> Oh! I see! that makes a lot more sense now.
> Is  there an  escape  sequence we  can use  instead  of writing  the
> unprintable characters into the file directly?

Seems there are two nice escaping modes for control-char

https://www.gnu.org/software/emacs/manual/html_node/elisp/Ctl_002dChar-Syntax.html#Ctl_002dChar-Syntax

The \C-something

(please  note  that   the  second  argument  is   an  arbitrary  ASCII
representation of the  control code, this to prevent  github eating it
as did in the code, also i do not think email play nice with non ASCII
caracters).

```
(string-match-p "\C-L" "^L") ; => 0

(string-match-p "\C-_" "^_") ; => 0

(string-match-p "\\(\C-L\\)?\C-_\\(\C-L\\)?$" "^L^_^L") ; => 0

(string-match-p "\\(\C-L\\)?\C-_\\(\C-L\\)?$" "^_^L")   ; => 0

(string-match-p "\\(\C-L\\)?\C-_\\(\C-L\\)?$" "^_")     ; => 0

(string-match-p "\\(\C-L\\)?\C-_\\(\C-L\\)?$" "^_^L")     ; => 0

(string-match-p "\\(\C-L\\)?\C-_\\(\C-L\\)?$" "^L^_^Lfoo")  ; => nil

(string-match-p "\\(\C-L\\)?\C-_\\(\C-L\\)?$" "foo^_^L")  ; => 3

(string-match-p "\\(\C-L\\)?\C-_\\(\C-L\\)?$" "foo^L^_^L")  ; => 3 ```

I tried the The "?\^-something" escaping but, sadly, i was not able to
use it in a regular expression. :(

Looking forward to know your opinion!

Bye!
C.

bastibe:
I think in this case, the `"\^L"` syntax is the most readable in this case.

The `"?\^L"` does not work, as `?` is the elisp literal for a single character: `?a` is a different way of writing `97`, and `?^L` is the same as `12`. The `?` syntax therefore doesn't work inside quotation marks.

cage2:
On Wed, Nov 27, 2019 at 12:46:32AM -0800, Bastian Bechtold wrote:

> I think  in this case,  the `"\^L"` syntax  is the most  readable in
> this case.
>
> The `"?\^L"` does not work, as `?` is the elisp literal for a single
> character: `?a` is a different way of writing `97`, and `?^L` is the
> same as `12`. The `?` syntax therefore doesn't work inside quotation
> marks.

Ahhhh i  think i  see now,  i totally  missed that  the '?..'   is for
characters! :/ Thank you for your clarification!

Now i realized that even the  parentesis could be useless so the whole
regex reduce to: "\^L?\^_\^L?\^J", right?

As usual is alway useful discuss things with other people, i was stuck
in a "mental" dead end.

Bye!
C.

bastibe:
> Now i realized that even the parentesis could be useless so the whole regex reduce to: "\^L?\^_\^L?\^J", right?

This seems correct. What an odd format.

1. optional formfeed
2. unit separator
3. optional formfeed
4. linefeed

The unit separator makes sense, as does the linefeed. But I particularly wonder about the reason for the double optional formfeeds. Perhaps to clear the screen at info page boundaries? There must have been a technical reason for those at the time. There's history in this little definition...

cage2:
On Wed, Nov 27, 2019 at 11:59:36PM -0800, Bastian Bechtold wrote:
> > Now i realized that even the parentesis could be useless so the whole regex reduce to: "\^L?\^_\^L?\^J", right?
>
> This seems correct. What an odd format.
>
> - optional formfeed
> - unit separator
> - optional formfeed
> - linefeed
>
> The unit separator makes sense, as does the linefeed. But I particularly wonder about the reason for the double optional formfeeds. Perhaps to clear the screen at info page boundaries? There must have been a technical reason for those at the time. There's history in this little definition...

Indeed! And  someway i feel that,  writing code for emacs  today, i am
paying homage to  all the venerable CS history  artifacts that peoples
behind MIT AI lab gave us many decades ago.

Bye!
C.

cage2:
On Wed, Nov 27, 2019 at 11:59:36PM -0800, Bastian Bechtold wrote:

Hello bastian!

>There's history in this little definition...

I think could be nice to put  here the wikipedia links to info format,
too:

https://en.wikipedia.org/wiki/Info_(Unix)

Bye!
C.

cage2:
Closing as #51 contains the same fix.

PR #53, opened 2019-12-02
 Added the chance to delete annotation and modify the annotated text from summary window;

cage2:
Hi!

also:
- trying to group function by scope;
- fixed a few typos;
- replaced some string literals with constants.

Bye!
C.

### DIFF
--- a/annotate.el
+++ b/annotate.el
@@ -1199,8 +1199,6 @@ annotation."
                                                                       rest-annotations)
                                                               checksum)))
                  (push new-record
-                       rest-annotations)
-                 (push new-record
                        rest-of-db))
              db-records))
        db-records))))
--- a/annotate.el
+++ b/annotate.el
@@ -1129,6 +1129,21 @@ annotation."
     (let ((print-length nil))
       (prin1 data (current-buffer)))))
 
+(cl-defmacro with-matching-annotation-fns ((filename
+                                            beginning
+                                            ending)
+                                           &body body)
+  "Anaphoric macro to build functions to find annotations"
+  `(let ((filename-match-p          (lambda (record)
+                                      (string= (annotate-filename-from-dump record)
+                                               ,filename)))
+         (annotation-limits-match-p (lambda (a)
+                                      (and (= (annotate-beginning-of-annotation a)
+                                              ,beginning)
+                                           (= (annotate-ending-of-annotation    a)
+                                              ,ending)))))
+     ,@body))
+
 (defun annotate-db-remove-annotation (db-records
                                       record-filename
                                       annotation-beginning
@@ -1136,67 +1151,59 @@ annotation."
   "Remove from database `db-records' the annotation indentified by
  the triplets `record-filename', `annotation-beginning' and
  `annotation-ending'; if such annotation does exists."
-  (let* ((filename-match-p (lambda (record)
-                             (string= (annotate-filename-from-dump record)
-                                      record-filename)))
-         (file-matched-record (cl-find-if filename-match-p db-records)))
-    (if file-matched-record
-        (let* ((rest-of-db      (cl-remove-if filename-match-p db-records))
-               (new-annotations (cl-remove-if (lambda (a)
-                                                (and (= (annotate-beginning-of-annotation a)
-                                                        annotation-beginning)
-                                                     (= (annotate-ending-of-annotation    a)
-                                                        annotation-ending)))
-                                              (annotate-annotations-from-dump file-matched-record)))
-               (checksum        (annotate-checksum-from-dump file-matched-record))
-               (new-record      (annotate-make-record record-filename
-                                                      new-annotations
-                                                      checksum)))
-          (push new-record
-                rest-of-db))
-      db-records)))
+  (with-matching-annotation-fns
+   (record-filename
+    annotation-beginning
+    annotation-ending)
+   (let ((file-matched-record (cl-find-if filename-match-p db-records)))
+     (if file-matched-record
+         (let* ((rest-of-db      (cl-remove-if filename-match-p db-records))
+                (new-annotations (cl-remove-if annotation-limits-match-p
+                                               (annotate-annotations-from-dump file-matched-record)))
+                (checksum        (annotate-checksum-from-dump file-matched-record))
+                (new-record      (annotate-make-record record-filename
+                                                       new-annotations
+                                                       checksum)))
+           (push new-record
+                 rest-of-db))
+      db-records))))
 
 (defun annotate-db-replace-annotation (db-records
                                        record-filename
                                        annotation-beginning
                                        annotation-ending
                                        replacing-text)
-  "TODO"
-  (let* ((filename-match-p (lambda (record)
-                             (string= (annotate-filename-from-dump record)
-                                      record-filename)))
-         (file-matched-record (cl-find-if filename-match-p db-records)))
-    (if file-matched-record
-        (let ((old-annotation   (cl-find-if (lambda (a)
-                                              (and (= (annotate-beginning-of-annotation a)
-                                                      annotation-beginning)
-                                                   (= (annotate-ending-of-annotation    a)
-                                                      annotation-ending)))
-                                            (annotate-annotations-from-dump file-matched-record))))
-             (if old-annotation
-                 (let* ((rest-of-db       (cl-remove-if filename-match-p db-records))
-                        (rest-annotations (cl-remove-if (lambda (a)
-                                                          (and (= (annotate-beginning-of-annotation a)
-                                                                  annotation-beginning)
-                                                               (= (annotate-ending-of-annotation    a)
-                                                                  annotation-ending)))
-                                                        (annotate-annotations-from-dump file-matched-record)))
-                        (checksum        (annotate-checksum-from-dump file-matched-record))
-                        (new-annotation  (annotate-make-annotation annotation-beginning
-                                                                   annotation-ending
-                                                                   replacing-text
-                                                                   (annotate-annotated-text old-annotation)))
-                        (new-record      (annotate-make-record record-filename
-                                                               (append (list new-annotation)
-                                                                       rest-annotations)
-                                                               checksum)))
-                   (push new-record
-                         rest-annotations)
-                   (push new-record
-                         rest-of-db))
-               db-records))
-      db-records)))
-
+  "Replace the text of annotation from database `db-records'
+ indentified by the triplets `record-filename',
+ `annotation-beginning' and `annotation-ending'; if such
+ annotation does exists."
+  (with-matching-annotation-fns
+   (record-filename
+    annotation-beginning
+    annotation-ending)
+   (let ((file-matched-record (cl-find-if filename-match-p db-records)))
+     (if file-matched-record
+         (let ((old-annotation   (cl-find-if annotation-limits-match-p
+                                             (annotate-annotations-from-dump file-matched-record))))
+           (if old-annotation
+               (let* ((rest-of-db       (cl-remove-if filename-match-p db-records))
+                      (rest-annotations (cl-remove-if annotation-limits-match-p
+                                                      (annotate-annotations-from-dump file-matched-record)))
+                      (checksum         (annotate-checksum-from-dump file-matched-record))
+                      (new-annotation   (annotate-make-annotation annotation-beginning
+                                                                  annotation-ending
+                                                                  replacing-text
+                                                                  (annotate-annotated-text old-annotation)))
+                      (new-record       (annotate-make-record record-filename
+                                                              (append (list new-annotation)
+                                                                      rest-annotations)
+                                                              checksum)))
+                 (push new-record
+                       rest-annotations)
+                 (push new-record
+                       rest-of-db))
+             db-records))
+       db-records))))
 
 ;;;; database related procedures ends here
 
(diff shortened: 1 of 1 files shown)


annotate.el
```diff
@@ -1073,6 +1090,116 @@ example:
...
+  (with-temp-file annotate-file
+    (let ((print-length nil))
+      (prin1 data (current-buffer)))))
+
+(defun annotate-db-remove-annotation (db-records
+                                      record-filename
+                                      annotation-beginning
+                                      annotation-ending)
+  "Remove from database `db-records' the annotation indentified by
+ the triplets `record-filename', `annotation-beginning' and
```
bastibe, line 1152: What happens if the triplet does not match? Can this happen?

annotate.el
```diff
@@ -1073,6 +1090,116 @@ example:
...
+ `annotation-ending'; if such annotation does exists."
+  (let* ((filename-match-p (lambda (record)
+                             (string= (annotate-filename-from-dump record)
+                                      record-filename)))
+         (file-matched-record (cl-find-if filename-match-p db-records)))
+    (if file-matched-record
+        (let* ((rest-of-db      (cl-remove-if filename-match-p db-records))
```
bastibe: I kind of dislike the use of `let*`. I often find myself using `let*` as a simple way of writing "procedural" Python code in lisp.

In this case, I would construct `rest-of-db` and `new-annotation` in a plain `let`, and inline `checksum` and `new-record` into `push`. But I honestly don't know if this is idiomatic or not.

annotate.el
```diff
@@ -1073,6 +1090,116 @@ example:
...
+          (push new-record
+                rest-of-db))
+      db-records)))
+
+(defun annotate-db-replace-annotation (db-records
+                                       record-filename
+                                       annotation-beginning
+                                       annotation-ending
+                                       replacing-text)
+  "TODO"
```
bastibe: Docstring missing (obviously)

annotate.el
```diff
@@ -1073,6 +1090,116 @@ example:
...
+  (let* ((filename-match-p (lambda (record)
+                             (string= (annotate-filename-from-dump record)
+                                      record-filename)))
+         (file-matched-record (cl-find-if filename-match-p db-records)))
+    (if file-matched-record
```
bastibe: In other programming languages, I would insert a `if not file-matched-record: return db-records` here, in order to reduce the nesting.

Perhaps this could be done in lisp as well by inlining the `if` into the `let`, i.e. `(let ((old-annotation (if file-matched-record (cl-find-if ...) nil))))`. Or even `(let ((old-annotation (and file-matched-record (cl-find-if ...)))))`.

Again, I don't know if this is more or less idiomatic, but at least it reduces some of the nesting, and possibly the trailing default return values.

annotate.el
```diff
@@ -1073,6 +1090,116 @@ example:
...
+                        (checksum        (annotate-checksum-from-dump file-matched-record))
+                        (new-annotation  (annotate-make-annotation annotation-beginning
+                                                                   annotation-ending
+                                                                   replacing-text
+                                                                   (annotate-annotated-text old-annotation)))
+                        (new-record      (annotate-make-record record-filename
+                                                               (append (list new-annotation)
+                                                                       rest-annotations)
+                                                               checksum)))
+                   (push new-record
```
bastibe: Why is `new-record` pushed on both the `rest-annotations` and `rest-of-db`? Or should the second `push` push `rest-annotations` onto `rest-of-db`? It seems to me that we are dropping `rest-annotations` here.

bastibe (changes requested):
Wow, another cool feature!

Could you post another screen shot of how the summary looks now?

Also, I would be grateful if you could split pull requests into multiple commits in the future. In particular, it is hard to judge from the diff which of the parts are actually new, vs the parts that were merely rearranged.

I added a few comments on programming style. I am not an Elisp expert at all, so take everything I say about elisp with a huge helping of salt. But at the very least I would be very interested in your take on these comments.

annotate.el
cage2: Hi Bastian!
Ouch! That form:

```
  (push new-record
                       rest-annotations)
```

was just wrong. :/
Bye!
C.

annotate.el
cage2: Hi!
Yes this is likely to happens!

If `file-matched-record` is nil the value of the function is just `db-records` (no changes, so far so good)

if `file-matched-record` is not nil (i.e. its value is the record that contains `record-filename`) and
there is no annotation in said record with limits equals to `annotation-beginning` and 
`annotation-ending` then:
- `rest-of-db` values the db without the record that match `record-filename` (and this record must exists as `file-matched-record` is non-nil);
- new-annotations contains exactly the same annotation as in the original record (because `cl-remove-if` is going to remove nothing).

so here:

``` lisp
(new-record  (annotate-make-record record-filename
                                   new-annotations
                                   checksum)))
```
`new-record` is equals to the original record with `record-filename` and

```
(push new-record
      rest-of-db))
```
the value of this form is equivalent to `db-records`, nothing changed actually.
If that there is no fallacy in that logic (nor really sure there was not :-) ) if there is no match the database returned is always at least equivalent to the one passed as argument.

Do you agree? Because i am not totally sure now that i wrote that! :-D :-D

Bye!
C.

annotate.el
cage2: >I kind of dislike the use of let*. I often find myself using let* as a simple way of writing "procedural" >Python code in lisp.

Indeed this (as far as i know!) is what is used for! :) Anyway consider that this form below:

``` lisp
(let* ((a 1)
       (b 2))
  (- a b))
```
is just syntactic sugar for:

``` lisp
(let ((a 1))
  (let ((b 2))
    (- a b)))
```
that, in turn, is just syntactic sugar for:

``` lisp
((lambda (a)
   ((lambda (b) (- a b))
    2))
 1)
```
So this is actually very lispy, i suppose! :D

annotate.el
cage2: Right, added!

annotate.el
cage2: Usually (but is not a rule cast in stone :-)) i prefer `(if something ...` to `(if (not something...`, just less cognitive burden for me (if i understood correctly what you wrote); if you prefer the form the `not` i could change that as the test is very simple (the simplest possible :) ).

I kind to avoid  complex forms when binding variable: but, again this are just stylistic preferences.

Bye!
C.

cage2:
> Wow, another cool feature!

Thank you!

> Could you post another screen shot of how the summary looks now?

Yes, of course:

https://www.autistici.org/interzona/img/screen-summary-mod.png

There is room for improvement, i guess! ;))

> Also, I would be grateful if you could split pull requests into multiple commits in the future. In particular, it is hard to judge from the diff which of the parts are actually new, vs the parts that were merely rearranged.

Sorry, this is just my fault. I made everything in a single commit and after i realized that this could made difficult to focus on the important modification: i has been lazy. :(

> 
> I added a few comments on programming style. I am not an Elisp expert at all, so take everything I say about elisp with a huge helping of salt. But at the very least I would be very interested in your take on these comments.

Bastian believe me, maybe (again, maybe) i have some more experience in another dialect of lisp (Common Lisp) but i never touched a single line of elisp before starting with this project so, from my point of view  you are the elisp expert of us. :) And i can guarantee that elisp's  defaulting to variable dynamic scoping was troublesome for some of my programming habits :-) (i still wonder if closures are possible, for example).

Bye!
C.

annotate.el
bastibe: Seems good, thank you for clarifying.

It *feels* like there's a cleaner way to express this. All the `find-if` and `remove-if` make my head spin a little. But maybe that's also just a consequence of the `cl-` functions available to us. Anyway, this looks good to me.

annotate.el
bastibe: Yeah, it's cool how lisp just decomposes into plain old lambdas. For some reason, I sometimes empathize with the machine, and try to nest as few parenthesis as possible; i.e. try to use `let` instead of `let*`, as in this case. But I guess that's actually silly :-)

annotate.el
bastibe: Interesting!

I think I was advocating that it might be better to push complexity into the variable binding to avoid complexity in the ensuing code. You are saying exactly the opposite.

That's fascinating. I see merits in both ways of looking at it.

I wish we could meet for a beer one day and just talk about programming for a while ;-).

bastibe:
> And i can guarantee that elisp's defaulting to variable dynamic scoping was troublesome for some of my programming habits :-)

As of Emacs 25 (I think), we can actually use lexical scope now, as a file-wide preference. And yes, I totally agree that dynamic scope threw me for a loop as well.

I like the macro you included recently. That's another language feature that took some getting used to. It simplifies these two functions nicely.

I have been thinking about how an Emacs would be designed today, though. And dynamic binding has certain advantages for self-modifying software in that every piece of the code can be swapped out at any time. Although I guess many programming languages are secretly dynamically scoped for global variables and global functions as well, so that helps somewhat.

Anyway, this looks good to me!

bastibe (approved):

cage2:
On Thu, Dec 05, 2019 at 11:54:36PM -0800, Bastian Bechtold wrote:

Hi Bastian!

> bastibe commented on this pull request.

[...]

> Seems good, thank you for clarifying.

My pleasure!

> It  *feels* like  there's a  cleaner way  to express  this.

Me too!

> All  the
> `find-if`  and `remove-if`  make my  head spin  a little.  But maybe
> that's also just  a consequence of the `cl-`  functions available to
> us. Anyway, this looks good to me.

My  concern is  that  the code  above  could be  (i  would say,  "is") slightly
inefficient. Anyway i think an optimization can be postponed.

Bye!
C.

cage2:
On Thu, Dec 05, 2019 at 11:59:03PM -0800, Bastian Bechtold wrote:

Hi!

> bastibe commented on this pull request.
>
>
>
> > +    (let ((print-length nil))
> +      (prin1 data (current-buffer)))))
> +
> +(defun annotate-db-remove-annotation (db-records
> +                                      record-filename
> +                                      annotation-beginning
> +                                      annotation-ending)
> +  "Remove from database `db-records' the annotation indentified by
> + the triplets `record-filename', `annotation-beginning' and
> + `annotation-ending'; if such annotation does exists."
> +  (let* ((filename-match-p (lambda (record)
> +                             (string= (annotate-filename-from-dump record)
> +                                      record-filename)))
> +         (file-matched-record (cl-find-if filename-match-p db-records)))
> +    (if file-matched-record
> +        (let* ((rest-of-db      (cl-remove-if filename-match-p db-records))
>
> Yeah, it's cool how lisp just decomposes into plain old lambdas.

Indeed! I remember the first time  a person shown me that (probably in
a book,  actually i do not  remember) i feel admiration  for the people
that designed the  language that (as one of its  dialect) we are still
using today.

> For some reason, I sometimes empathize  with the machine, and try to
>nest as few parenthesis as possible; i.e. try to use `let` instead of
>`let*`, as in this case. But I guess that's actually silly :-)

I don think it  is silly, instead i think that as  human being we just
can not  prevent ourselves to pour  our taste/preferneces/feeling into
our intellectual production no matter the nature of the works painting
or source code.

Maybe i am getting too philosophical now! :-D

Bye!
C.

cage2:
On Fri, Dec 06, 2019 at 12:04:19AM -0800, Bastian Bechtold wrote:

Hi!

[...]

> Interesting!
>
> I think I was advocating that it might be better to push complexity into the variable binding to avoid complexity in the ensuing code. You are saying exactly the opposite.
>
> That's fascinating. I see merits in both ways of looking at it.

As in a previous message, we are seeing the people differences mutated
into code style differences! That's fine! :)

> I wish we could meet for a beer one day and just talk about programming for a while ;-).

Would be my pleasure, to offer you a glass of beer! :)

There will  be the  european lisp  symposium in  Zurich next  April, i
would like to  go but yet i  did not make the  decision. Just throwing
the idea here!  :)

Bye!
C.

cage2:
On Fri, Dec 06, 2019 at 12:10:50AM -0800, Bastian Bechtold wrote:

Hi!!

> > And i can guarantee that elisp's defaulting to variable dynamic scoping was troublesome for some of my programming habits :-)

> As of Emacs 25 (I think), we  can actually use lexical scope now, as
> a file-wide preference. And yes,  I totally agree that dynamic scope
> threw me for a loop as well.

I would like to  enable it but i fear for regression so  i think i can
deal with it. :)


> I like the macro you included recently. That's another language feature that took some getting used to. It simplifies these two functions nicely.

Macro are cools! :)

> I have  been thinking about  how an  Emacs would be  designed today,
> though.   And   dynamic   binding   has   certain   advantages   for
> self-modifying  software in  that every  piece  of the  code can  be
> swapped out at any time. Although I guess many programming languages
> are  secretly dynamically  scoped  for global  variables and  global
> functions as well, so that helps somewhat.

In  common lisp  a  variable can  be specified  to  be dynamically  or
lexical scoped with a special form.   This is a powerful feature but
also can be a source of confusion, much more than dynamic scoping only;
in fact there is a convention to  wrap a special variable with '*' for
dynamic  variables.  Some  compilers  goes  so far  that  even show  a
warning if you do binding of  a dynamic variable without the asterisks.
And  i still  today i  thanks  the developers  that implemented  this
feature! :-D :-D

>
> Anyway, this looks good to me!

Very well! Merging!

Bye!
C.

Issue #20, opened 2015-09-23
Make position of annotations customizable


bastibe:
Annotations should go to either the right margin, or the next line.

Also, annotations of long lines should probably always go to the next line.

cage2:
fixed in #54

PR #54, opened 2019-12-06
Annotations on their own line

cage2:
Hi!

with this PR i try to add the feature as described in #20 

There is a variable `annotate-annotation-position-policy` that can specify the policy to set the annotation position: on a new line, on the margin or on a new line only when the annotation's size exceed the value of `annotate-annotation-max-size-not-place-new-line`.

I strongly feel that the name of the above variables can be changed in something better! :)

Bye!
C.

### DIFF
--- a/annotate.el
+++ b/annotate.el
@@ -140,6 +140,13 @@ database is not filtered at all."
   :type 'boolean
   :group 'annotate)
 
+(defcustom annotate-annotation-max-size-not-place-new-line 15
+ "The maximum `string-width` allowed for an annotation to be
+ placed on the right margin of the window instead of its own line
+ after the annotated text."
+  :type 'number
+  :group 'annotate)
+
 (defconst annotate-warn-file-changed-control-string
   (concat "The file '%s' has changed on disk "
           "from the last time the annotations were saved.\n"
@@ -756,12 +763,13 @@ to 'maximum-width'."
     (let ((newline-position (point)))
       (goto-char (1- (point))) ; we start at the start of the previous line
       ;; find overlays in the preceding line
-      (let ((prefix             (annotate-make-prefix)) ; white spaces before first annotation
-            (bol                (progn (beginning-of-line) (point)))
-            (eol                (progn (end-of-line) (point)))
-            (text               "")
-            (overlays           nil)
-            (annotation-counter 1))
+      (let* ((prefix-first       (annotate-make-prefix)) ; white spaces before first line of annotation
+             (prefix-rest        (make-string annotate-annotation-column ? ))
+             (bol                (progn (beginning-of-line) (point)))
+             (eol                (progn (end-of-line) (point)))
+             (text               "")
+             (overlays           nil)
+             (annotation-counter 1))
         ;; include previous line if point is at bol:
         (when (null (overlays-in bol eol))
           (setq bol (1- bol)))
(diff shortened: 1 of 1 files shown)


annotate.el
```diff
@@ -782,30 +802,54 @@ to 'maximum-width'."
...
+                                         annotate-annotation-max-size-not-place-new-line))
+                 (position            annotate-annotation-position-policy)
+                 (position-new-line-p (cl-case position
+                                        (:new-line
+                                         t)
+                                        (:by-length
+                                         annotation-long-p)
+                                        (otherwise
+                                         nil)))
+                 (splitted-annotation (if position-new-line-p
```
bastibe: Should be `split-annotation` I think. Or maybe `multiline-annotation` is better.

annotate.el
```diff
@@ -782,30 +802,54 @@ to 'maximum-width'."
                                                   (<= (overlay-start a)
                                                       newline-position
                                                       (overlay-end a))))
-                                      (overlays-in bol eol))
+                                  (overlays-in bol eol))
                     (lambda (x y)
                       (< (overlay-end x) (overlay-end y)))))
         ;; put each annotation on its own line
```
bastibe: This comment is now outdated and should be expanded.

annotate.el
```diff
@@ -140,6 +140,25 @@ database is not filtered at all."
...
+  "policy for annotation's position:
+  - :newline
+    always in a new-line
+  - :margin
+     always on right margin
+  - :by-length
+    decide by text's length
+"
+  :type  'symbol
+  :group 'annotate)
```
bastibe, line 165: This is a beautiful solution!

annotate.el
```diff
@@ -756,12 +775,13 @@ to 'maximum-width'."
     (let ((newline-position (point)))
       (goto-char (1- (point))) ; we start at the start of the previous line
       ;; find overlays in the preceding line
-      (let ((prefix             (annotate-make-prefix)) ; white spaces before first annotation
-            (bol                (progn (beginning-of-line) (point)))
-            (eol                (progn (end-of-line) (point)))
-            (text               "")
-            (overlays           nil)
-            (annotation-counter 1))
+      (let* ((prefix-first       (annotate-make-prefix)) ; white spaces before first line of annotation
```
bastibe: If I'm reading this correctly, this doesn't need to be a `let*` any longer, and could be a `let` instead.

annotate.el
```diff
@@ -782,30 +802,54 @@ to 'maximum-width'."
...
-                                  'annotate-highlight-secondary)))
+          (let* ((face                (if (= (cl-rem annotation-counter 2) 0)
+                                          'annotate-annotation
+                                        'annotate-annotation-secondary))
+                 (face-highlight      (if (= (cl-rem annotation-counter 2) 0)
+                                          'annotate-highlight
+                                        'annotate-highlight-secondary))
+                 (annotation-long-p   (> (string-width (overlay-get ov 'annotation))
+                                         annotate-annotation-max-size-not-place-new-line))
+                 (position            annotate-annotation-position-policy)
```
bastibe: Why rename the existing variable?

annotate.el
```diff
@@ -782,30 +802,54 @@ to 'maximum-width'."
...
+            (cl-incf annotation-counter)
             (overlay-put ov 'face face-highlight)
-            (dolist (l (save-match-data
-                         (split-string (annotate-lineate (overlay-get ov 'annotation)
-                                                         (- eol bol))
-                                       "\n")))
+            (when position-new-line-p
+              (setf prefix-first " \n"))
+            (dolist (l splitted-annotation)
               (setq text
```
bastibe: It feels like `text` is to generic a variable name for its scope. I think I'd either make `text` much more local, or give it a more descriptive name. (In general I try to use short names only very locally, where their meaning is clear from context).

bastibe (changes requested):
Really cool! I love the way you implemented the choices, particularly with the default being an "intelligent" choice between the two.

As usual, find attached a few comments. Looks really good already, though!

I find it interesting how the first `let` in this function is mostly for placeholders that are filled with content later (C style), while the second `let` is more of a procedural construction of complex data from simpler data. I wonder which style is more idiomatic in elisp.

cage2:
Hei Bastian!

> Really cool! I love the way you implemented the choices, particularly with the default being an "intelligent" choice between the two.

Thank you! Honestly that way to manage the position of the annotation came mostly natural because the  code for this feature was more or less already there, as you can see the lines added to implement this feature are not a lot and i like to consider this this like a measure that the package's code is not too bad. ;-) ;-)

> 
> As usual, find attached a few comments.

OK, i tried to address them in the just pushed commits.

> Looks really good already, though!

great! :+1: 
 
> I find it interesting how the first `let` in this function is mostly for placeholders that are filled with content later (C style), while the second `let` is more of a procedural construction of complex data from simpler data. 

This is an interesting consideration! I found nice that you can mix two different style of programming so seamless! 

> I wonder which style is more idiomatic in elisp.
Me too!

Bye!  
C.

> Really cool! I love the way you implemented the choices, particularly with the default being an "intelligent" choice between the two.
> 
> As usual, find attached a few comments. Looks really good already, though!
> 
> I find it interesting how the first `let` in this function is mostly for placeholders that are filled with content later (C style), while the second `let` is more of a procedural construction of complex data from simpler data. I wonder which style is more idiomatic in elisp.

annotate.el

Right, i have changed it to `multiline-annotation`

Bye!
C.

annotate.el

Yes, indeed it did not made much sense, i have expanded a bit. Should the rest of the code inside the annotations iteration `(dolist (ov overlays) ...)` needs comments too?

Bye!
C.

annotate.el
cage2: Thank you! :)

annotate.el

Yes there is no need for `let*` and moreover i think if we left it there this could be confusing for people that read the source, creating false expectations: that a previous binding is used in one of the next when truly is not.

Bye!
C.

annotate.el

No reason at all, just left there from a refactoring and i forgot to remove it. :(

Bye!
C.

annotate.el

I agree text is a bit generic, so I changed it with `annotated-text` t, is it OK for you?

Bye!!
C.

bastibe (approved):

bastibe:
Everything looks good to me!

Issue #59, opened 2020-02-20
Rethinking multiline annotation


cage2:
Hello!

There are currently issues with multi line annotations, for helping discussion i paste here the relevant comment in PR #58 

https://github.com/bastibe/annotate.el/pull/58#issuecomment-583833587

> Hi Bastian!

> 
> I was investigating the issue we got annotating the newline and i have met some bugs in the process! :)
> 
> Unfortunately i come to the conclusion that is not possible to annotate the newline itself because of the role if has in the annotation process, in fact, when annotated, it propagate the annotation overlay (the one with the underline property) to the prefix text we use to pad the annotation (the actual text of annotation, i mean). This is a problem not only when try to annotate the newline but, also, when the annotated text is a multi line one (using a region) and -of course- it include a newline char.
> 
> A picture, probably is going to explain better:
> 
> https://www.autistici.org/interzona/img/misc/annot-long-overlay-newline.png
> 
> I scratched my head about this problem, i think i could split the multi line annotation in multiple overlay that skips the newline character: there will be a lot of works to refactor all the functionality but i hope this is doable, at least; i can not see solution when the only character annotated is the newline.
> 
> I hope there is a proper solution, though. Like if was possible to delete the overlay on part of the text or just remove the underline property.
> 
> Bye!
> C.

I have a local branch that try to address this issue and my limited testing seems to shows that works but, even if everything worked (and i doubt this was true) this is just a starting point, comments are welcome!
Bye!
C.

PS: I will  post a PR  with the mentioned branch in the next hours.

cage2:
Hello!

I just pushed a branch with proposed changes to fix the multi line annotation bug.

The main idea is that an annotation can be decomposed in a "chain" of overlays: each overlay, in a chain, is identified by a number property that starts from 0 (for the leftmost overlay) and increase going right (further in the buffer), the last overlay of an annotation is marked with a negative number.

Looking forward for comments! 

bye!
C.

PR #60, opened 2020-02-21
Rethink multiline annotations

cage2:
See : #59

### DIFF
--- a/annotate.el
+++ b/annotate.el
@@ -69,9 +69,9 @@ See https://github.com/bastibe/annotate.el/ for documentation."
 
 (define-key annotate-mode-map (kbd "C-c C-s") 'annotate-show-annotation-summary)
 
-(define-key annotate-mode-map (kbd "C-c ]") 'annotate-next-annotation)
+(define-key annotate-mode-map (kbd "C-c ]") 'annotate-move-next-annotation)
 
-(define-key annotate-mode-map (kbd "C-c [") 'annotate-previous-annotation)
+(define-key annotate-mode-map (kbd "C-c [") 'annotate-move-previous-annotation)
 
 (defcustom annotate-file (locate-user-emacs-file "annotations" ".annotations")
   "File where annotations are stored."
@@ -164,6 +164,15 @@ database is not filtered at all."
 "
   :type  'symbol
   :group 'annotate)
+
+(defconst annotate-prop-chain-position
+  'position)
+
+(defconst annotate-prop-chain-pos-marker-first
+  0)
+
+(defconst annotate-prop-chain-pos-marker-last
+  -1)
 
 (defconst annotate-warn-file-changed-control-string
   (concat "The file '%s' has changed on disk "
@@ -367,7 +376,7 @@ modified (for example a newline is inserted)."
         (create-new-annotation)))
       (set-buffer-modified-p t))))
 
-(defun annotate-next-annotation ()
+(defun annotate-move-next-annotation ()
   "Move point to the next annotation."
   (interactive)
   ;; get all following overlays
@@ -387,7 +396,7 @@ modified (for example a newline is inserted)."
       ;; jump to first overlay list
       (goto-char (overlay-start (nth 0 overlays))))))
 
-(defun annotate-previous-annotation ()
+(defun annotate-move-previous-annotation ()
   "Move point to the previous annotation."
   (interactive)
   ;; get all previous overlays
@@ -807,23 +816,8 @@ to 'maximum-width'."
         (when (null (overlays-in bol eol))
           (setq bol (1- bol)))
         (setq overlays
-              (sort (cl-remove-if (lambda (a) (or (not (annotationp a))
-                                                  ;; if an annotated
-                                                  ;; text contains a
-                                                  ;; newline (is a
-                                                  ;; multiline one) do
-                                                  ;; not add
-                                                  ;; annotation for it
-                                                  ;; here (i.e. remove
-                                                  ;; from that list),
-                                                  ;; this annotation
-                                                  ;; will be shown on
-                                                  ;; the next newline
-                                                  ;; instead
-                                                  (<= (overlay-start a)
-                                                      newline-position
-                                                      (overlay-end a))))
-                                  (overlays-in bol eol))
+              (sort (cl-remove-if-not #'annotationp
+                                      (overlays-in bol eol))
                     (lambda (x y)
                       (< (overlay-end x) (overlay-end y)))))
         ;; configure each annotation's properties and place it on the
@@ -831,9 +825,15 @@ to 'maximum-width'."
         ;; or right marigin) is indicated by the value of the
         ;; variable: `annotate-annotation-position-policy'.
         (dolist (ov overlays)
-          (let* ((face                (if (= (cl-rem annotation-counter 2) 0)
-                                          'annotate-annotation
-                                        'annotate-annotation-secondary))
+          (let* ((face                (cond
+                                       ((not (annotate-chain-first-p ov))
+                                        (let ((first-in-chain (annotate-chain-first ov)))
+                                          (overlay-get first-in-chain
+                                                       'annotation-face)))
+                                       ((= (cl-rem annotation-counter 2) 0)
+                                        'annotate-annotation)
+                                       (t
+                                        'annotate-annotation-secondary)))
                  (face-highlight      (if (= (cl-rem annotation-counter 2) 0)
                                           'annotate-highlight
                                         'annotate-highlight-secondary))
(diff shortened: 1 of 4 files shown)


annotate.el
```diff
@@ -69,9 +72,9 @@ See https://github.com/bastibe/annotate.el/ for documentation."
 
 (define-key annotate-mode-map (kbd "C-c C-s") 'annotate-show-annotation-summary)
 
-(define-key annotate-mode-map (kbd "C-c ]") 'annotate-next-annotation)
+(define-key annotate-mode-map (kbd "C-c ]") 'annotate-move-next-annotation)
```
bastibe: I think "move" is ambiguous, as it might be interpreted as moving the annotation. Perhaps "goto" is better?
```suggestion
(define-key annotate-mode-map (kbd "C-c ]") 'annotate-goto-next-annotation)
```

annotate.el
```diff
@@ -69,9 +72,9 @@ See https://github.com/bastibe/annotate.el/ for documentation."
...
 
-(define-key annotate-mode-map (kbd "C-c [") 'annotate-previous-annotation)
+(define-key annotate-mode-map (kbd "C-c [") 'annotate-move-previous-annotation)
```
bastibe: See above
```suggestion
(define-key annotate-mode-map (kbd "C-c [") 'annotate-goto-previous-annotation)
```

annotate.el
```diff
@@ -367,7 +376,7 @@ modified (for example a newline is inserted)."
         (create-new-annotation)))
       (set-buffer-modified-p t))))
 
-(defun annotate-next-annotation ()
+(defun annotate-move-next-annotation ()
```
bastibe:
```suggestion
(defun annotate-goto-next-annotation ()
```

annotate.el
```diff
@@ -387,7 +396,7 @@ modified (for example a newline is inserted)."
       ;; jump to first overlay list
       (goto-char (overlay-start (nth 0 overlays))))))
 
-(defun annotate-previous-annotation ()
+(defun annotate-move-previous-annotation ()
```
bastibe:
```suggestion
(defun annotate-goto-previous-annotation ()
```

bastibe (commented):
Wow, another big change!

In general, this looks good. What I don't understand yet: What annotation-prop-chain-position is assigned to single-line annotations?

In other words, are there now two "classes" of annotations, one a chain and one a single annotation, that are handled differently by the code, or is a single-line annotation simply a chain of length 1?

Otherwise, I applaud your efforts, but haven't been able to follow all parts of the code. A few comments would probably help me a lot!

bastibe:
Also, it occurred to me that it might be possible to manually set the `underline` face of annotations to `nil` instead of having to split the overlays. Or possibly, add a single-character overlay with a `nil` underline on the newline. Did you check if something like that is possible?

annotate.el
cage2: Hello!

I agree it is better to rename, this is a case where 'goto' is not considered harmful! ;-D

cage2:
Hello Bastian!

> Also, it occurred to me that it might be possible to manually set the `underline` face of annotations to `nil` instead of having to split the overlays. Or possibly, add a single-character overlay with a `nil` underline on the newline. Did you check if something like that is possible?

I tried the first with no luck but the second seems a good point, let me explore this idea! 

Bye!
C.

cage2:
Hi!

> Wow, another big change!

Thank you! :)
 
> In general, this looks good. What I don't understand yet: What annotation-prop-chain-position is assigned to single-line annotations?
> 
> In other words, are there now two "classes" of annotations, one a chain and one a single annotation, that are handled differently by the code, or is a single-line annotation simply a chain of length 1?

the second, i assigned the position `-1` to a chain made of single annotation, therefore the member of this this particular class of "chains" is both the first and the last. Seems the right thing to do but i fear this was quite arbitrary in retrospective, what do you think?

> 
> Otherwise, I applaud your efforts, but haven't been able to follow all parts of the code. A few comments would probably help me a lot!

Sure, let me add some more comments!

Bye!
C.

bastibe:
Cool! That makes sense, then.

For some reason, the numbering and specifically the `-1` feels off to me. But I can't quite put my finger on the reason, or a better solution. Maybe `nil` or a dedicated symbol might be better than `-1`? But then that's just playing games with syntax, and not really meaningful. I wonder if there is a different way, perhaps by linking the overlays explicitly with a linked list, where each overlay refers to the next or previous overlay in a property or something like that.

Anyway, these are just idle thoughts, and not important unless you find the topic interesting as a coding puzzle.

cage2:
On Tue, Mar 03, 2020 at 11:52:10PM -0800, Bastian Bechtold wrote:

Hello Bastian!

> Cool! That makes sense, then.

Well i  feel somehow is  a bit  "hacky" but at  least it does  fix the
problem and, with my surprise it the fact that an annotation (a single
overlay,  i mean)  can lay  on a  single line  made some  code simpler
(look, for example  at the diff generating code). And  this was totally
unintentional of course! ;-D ;-D

> For some reason,  the numbering and specifically the  `-1` feels off
> to me. But  I can't quite put  my finger on the reason,  or a better
> solution. Maybe  `nil` or  a dedicated symbol  might be  better than
> `-1`?

I do not like 'nil' because (if i remember correctly) if a property of
annotation does  not exists the expression  (overlay-get-property ...)
values  nil as  well,  moreover i  prefer  to keep  the  value of  the
properties of  the same type,  but these,  as you wrote,  are actually
matter of taste! :)

> But  then that's  just playing  games  with syntax,  and not  really
> meaningful. I wonder if there is a different way, perhaps by linking
> the  overlays explicitly  with  a linked  list,  where each  overlay
> refers to  the next or previous  overlay in a property  or something
> like that.

This  would be  a better  (or at  least cleaner)  solution, but  i was
scared to use buffer local variables and other thing i never used :)

> Anyway, these are  just idle thoughts, and not  important unless you
> find the topic interesting as a coding puzzle.

Probably  using an  external data  structure would  be interesting,  i
would   open   an   issue    just   as   remainder   mentioning   this
improvement. Maybe this could attract  to the project some other coder,
you never can tell! :)

Bye!
C.

bastibe:
True, I hadn't thought of that `nil` problem. And the linked list proposal feels like it could break easily.

>  i would   open   an   issue    just   as   remainder   mentioning   this improvement. Maybe this could attract  to the project some other coder, you never can tell! :)

Alright!

cage2:
Hello Bastian! 
I locally merged the master branch into this one, things seems to works so far but i would like to spend a few hours using the package on an actual code base (mine, so if i break things nobody will be upset, except me! ;-) )

Bye!
C.

cage2:
Hi Bastian!

After last fix I think i am more or less ready to merge this :)

Bye!
C.

bastibe:
OK, cool!

Although we should probably increment the minor version number instead of the bugfix number, as these changes are more akin to new features than to bug fixes. (In fact, we should probably think about releasing a new major version soon, as there were a few backwards-incompatible changes.)

cage2:
On Tue, Mar 10, 2020 at 12:41:02AM -0700, Bastian Bechtold wrote:

Hello Bastian!

> OK, cool!

> Although  we  should probably  increment  the  minor version  number
> instead of the bugfix number, as  these changes are more akin to new
> features than to bug fixes.

I totally agree, should we number this version 0.6.0?

> (In fact, we should probably think about
> releasing  a   new  major  version   soon,  as  there  were   a  few
> backwards-incompatible changes.)

Do you think is time for version 1.0?

Bye!
C.

bastibe:
> I totally agree, should we number this version 0.6.0?

Yes, I think that's a good idea!

> Do you think is time for version 1.0?

I'll leave that up to you, honestly. I'd wait with 1.0 until you're happy with the state of the library and have implemented all of your planned big changes. At any rate, I would let the project rest for a month or so, and see how the current state works in day-to-day use before changing the major version.

cage2:
On Wed, Mar 11, 2020 at 02:51:08AM -0700, Bastian Bechtold wrote:

Hello Bastian!

> > I totally agree, should we number this version 0.6.0?
>
> Yes, I think that's a good idea!

Ok!

> > Do you think is time for version 1.0?
>
> I'll leave that up to you, honestly.

Wow, that is a big responsability! Thank  you for the trust, i hope it
will be deserved! :)

> I'd wait with  1.0 until you're happy with the  state of the library
> and have implemented all of your planned big changes. At any rate, I
> would  let the  project rest  for a  month or  so, and  see how  the
> current  state works  in day-to-day  use before  changing the  major
> version.

The package appears more or less  feature-complete to me now, but i am
quite sure  some bugs (hopefully a  small number) are going  to pop-up
soon,  given  that  the  changes  that this  branch  brings  are  very
pervasive.  So, basically i agree to pospone the 1.0 to the future for
at least a  month with no bugs reports. In  the meanwhile, let's merge
this branch and let the code runs!  :)

Bye!
C.

PR #61, opened 2020-02-21
Org mode partial fix

cage2:
Hello Bastian!
This branch was sitting on my hard disk for a long time, unfortunately annotating ORG buffers still does not works for source block (#BEGIN_SRC ... #END_SRC) but otherwise seems ORG file can be annotated.

There are a couple more of bugfix in this PR so we can just choose to ignore the org-mode patch and just merge the other fix. 

Bye!
C.

### DIFF
--- a/Changelog
+++ b/Changelog
@@ -27,7 +27,7 @@
         disappears).
 
 2020-02-10 Bastian Bechtold, cage
-	* annotate.el (annotate--font-lock-matcher annotate-bounds nnotate-symbol-strictly-at-point annotate-next-annotation-change annotate-previous-annotation-change annotate-clear-annotations annotate-annotate)
+	* annotate.el (annotate--font-lock-matcher annotate-bounds annotate-symbol-strictly-at-point annotate-next-annotation-change annotate-previous-annotation-change annotate-clear-annotations annotate-annotate)
 	- prevented fails of fontification of annotated regions
 	As we walk across the overlays we can get past the limit;
 	- mark buffer as modified even if the only action the user performed
@@ -35,3 +35,17 @@
 	the file)
 	- prevented annotation of text marked with a region that overlap with
 	an existing annotation.
+
+2020-03-06 Bastian Bechtold, cage ::
+	* annotate.el (annotate-annotation-force-newline-policy annotate-annotation-newline-policy-forced-p annotate-summary-delete-annotation-button-pressed annotate--annotation-builder)
+
+	- used  an heuristic  to force newline  policy when  the annotated
+	text  does  not  uses a  standard  fonts  (using  font  height  as
+	comparison);
+
+	-  when, in  summary  window,  the delete  button  is pressed  the
+	software take care of reload  annotate mode for the visited buffer
+	the annotation button is referring to;
+
+	- when re-flowing annotation the window width was calculated always
+	for the current buffer (the one with the focus).
--- a/Changelog
+++ b/Changelog
@@ -1,3 +1,21 @@
+2020-01-25 Bastian Bechtold, cage
+
+        * annotate.el (defun annotate-annotation-force-newline-policy,
+                       annotate-annotation-newline-policy-forced-p,
+		       annotate-create-annotation,
+                       annotate-lineate,
+                       annotate-summary-delete-annotation-button-pressed):
+	- mitigated bug that prevented rendering of annotation in
+          org-mode forcing 'newline' policy for annotation
+          positioning.
+          See the local function
+          'maybe-force-newline-policy' in 'annotate-create-annotation'.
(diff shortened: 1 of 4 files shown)


bastibe:
Looks good as far as I can tell. 

Could you elaborate on the problem with the newline policy? It seems to be related to line-height somehow, but I couldn't quite figure out what was changed.

cage2:
On Mon, Mar 02, 2020 at 02:19:53AM -0800, Bastian Bechtold wrote:

Hi Bastian!

> Looks good as far as I can tell.  Could you elaborate on the problem
> with  the newline  policy?

yes, of course!

> It  seems  to be  related to  line-height
> somehow, but I couldn't quite figure out what was changed.

Uh,  this  is  really  strange...So basically  i  stumbled  upon  this
problem:  when the  annotated text  is  not of  the same  font as  the
annotation the "padding"  text (the spaces that  pushes the annotation
on the  right side of the  window) inherits the face  of the annotated
text and, even if the "padding" length (in column unit) is correct the
size of  the same number of  columns in pixel size  could differs, the
net result is that the annotation  would be moved away from the window
border even making it disappear.

So i  tried to force a  default face to annotation  (and padding) text
and this works; but turned out that,  if the font of annotated text is
big enough (in pixel units), even in that case the annotation could be
pushed away beyond the window's border.

This is  not limited  to org-mode but  i use non  default face,  on my
system, only in this mode. :)

So said,  the best solution i  found was to force  'newline' policy if
the font of annotated text is different from default one, why i chosen
the height of the font, actually i do know! O_O

Probably i  could chose  the width  not the height,  i think  i should
check if there is a valid  reason why i chose this particular criteria
(height). Probably  i will discover  that there  was not such  a valid
reason! :D

Bye!
C.

PS: i  am quite  sure i made  confusion between font  and face  in the
message above, sorry but font is difficult in Emacs! :)

bastibe:
Thank you, I understand it better now. My default theme actually uses multiple fonts, so I am very familiar with these sorts of issues. I try to use fonts with compatible widths, at least, but line height is the bane of my multi-font color scheme... (referring to "font" here, as in the font file, or font-family, as Emacs calls it in its face-specs, whereas Emacs' "face" is more like a combination of font variants, decorations, and a font. I digress.)

This makes a lot of sense. If in doubt, force newline-policy. This definitely needs a mention in the README, but is a perfectly acceptable workaround in my book.

If I remember correctly, there is a function that determines the pixel-width of a piece of text in Emacs, which could be used for a "correct" solution. But this gets complicated *very* quickly, and messes up lineation in all kinds of creative ways. I seem to remember that the most important reason for Emacs being slow for long lines is that line breaking is crazy difficult with variable-width fonts, combining characters, and complex kerning rules.

cage2:
On Wed, Mar 04, 2020 at 12:03:56AM -0800, Bastian Bechtold wrote:

Hello Bastian!

> Thank you, I understand it better now. My default theme actually uses multiple fonts,

Yes, i think i read the post in your blog! :)

> so I am very familiar with these sorts of issues. I try to use fonts with compatible widths, at least, but line height is the bane of my multi-font color scheme... (referring to "font" here, as in the font file, or font-family, as Emacs calls it in its face-specs, whereas Emacs' "face" is more like a combination of font variants, decorations, and a font. I digress.)
>
> This makes a lot of sense. If in doubt, force newline-policy. This definitely needs a mention in the README, but is a perfectly acceptable workaround in my book.

Sure, i  am going  to add that,  should that goes  to the  known "bug
section"  or  another  one  with  title  like  "Unusual  but  intended
behaviour"?

So, after fixing the README i would like to check if everything is OK a
last time (changelog  and NEWS) and then i think  we could merge this!
:)


Bye!
C.

bastibe:
> Yes, i think i read the post in your blog! :)

So I do have readers after all. It's always interesting hearing from readers, as I do not have any kind of logging on the blog, and do not have any kinds of statistics at all.

> Sure, i  am going  to add that,  should that goes  to the  known "bug section"  or  another  one  with  title  like  "Unusual  but  intended behaviour"?

Maybe an FAQ section would be appropriate for this.

Merge when you're ready!

cage2:
Hello Bastian!

> > Yes, i think i read the post in your blog! :)
> 
> So I do have readers after all. It's always interesting hearing from readers,

I also liked the posts about Fourier transform and signal processing! :)

> as I do not have any kind of logging on the blog, and do not have any kinds of statistics at all.

And i appreciate that, also! :)
> 
> > Sure, i  am going  to add that,  should that goes  to the  known "bug section"  or  another  one  with  title  like  "Unusual  but  intended behaviour"?
> 
> Maybe an FAQ section would be appropriate for this.

Sure!
 
> Merge when you're ready!

Ok!
Bye!
C.

PR #62, opened 2020-03-16
Fixed annotation picking in 'annotate-annotate'

cage2:
Hello Bastian!
The fix proposed is simple fix but the bug was bad! ;)

Bye!
C. 

-----
At the beginning of the function  mentioned above we was picking the
first available  overlay.  But if the annotated  text contained multiple
overlays (at least one more of the annotation) and the annotation was  not the first we missed the overlay we was looking for (i.e. the annotation).

This means, for example, that the annotation was not modifiable.

edit: rephrased the fix description.

### DIFF
--- a/Changelog
+++ b/Changelog
@@ -10,7 +10,7 @@
           positioning.
           See the local function
           'maybe-force-newline-policy' in 'annotate-create-annotation'.
-        - choosen the window that contains the current buffer when resizing the annotations
+        - chosen the window that contains the current buffer when resizing the annotations
           see variable 'current-window' in 'annotate-lineate';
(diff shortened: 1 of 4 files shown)


bastibe:
Good find!

Always a good thing to let things sit for a while, and let bugs surface ;-)

cage2:
On Tue, Mar 17, 2020 at 02:33:11AM -0700, Bastian Bechtold wrote:

Hello Bastian!

> Good find!

Only casually found, to be honest! :D

> Always a good thing to let things sit for a while, and let bugs surface ;-)

Indeed! :)

I wont be surprised if more to  come in the future, but time will tell! :)

Bye!
C.

Issue #68, opened 2020-05-28
Feature Request: Keep annoations seperate from branch to branch


gopar:
Hey!
Thanks for the awesome package :) 

Would it be possible to keep annotations from one git branch separate from another branch?

cage2:
On Thu, May 28, 2020 at 03:10:49PM -0700, Daniel Gopar wrote:
> Hey!

Hi! :)

> Thanks for the awesome package :)

Thank you, happy that this is useful to you :)

> Would  it  be possible  to  keep  annotations  from one  git  branch
> separate from another branch?

So,  if  I can  understand  correctly,  you  would  like to  have  the
possibility to  load different  database of annotations  depending on
the branch of git repository, right?

If  yes, unfortunately  this  package  is not  aware  of  all the  git
machinery  at all;  plus it  loads a  single database  from a  single,
shared, file.

Probably we could works on  changing the database file "on-the-fly" (i
do  know if  this could  be easily  accomplished, though)  but in  the
meanwhile  - as  a workaround  -  you could  just add  to the  working
directory  of your  git project  a symlink  to a  file to  be used  as
database, making git  to tracks this symlink and  instructing Emacs to
load this new file as annotation's database.

Please, checkout this answer:

https://stackoverflow.com/questions/9556807/track-files-locally-but-never-allow-them-to-be-pushed-to-the-remote-repository/10490318#10490318

path/to/your/repo/ $ ln -s /path/to/annotation.db annotation.db

git add annotation.db and, then, commit the repo

After your symlink is added to repository

 just customize the `Annotate File` value:

M-x customize-group <RET> annotate

to point to your database symlink

to  change database  with  branch just  make the  symlink  point to  a
different file for each branch.

Using symlinks  prevents the accidental  sharing of the  database and,
depending of your needs, this could be an acceptable solutions.

I hope also this is an acceptable answer to your question! :)

Bye!
C.

PS: Of course  you can add the  database file to the  tracked files by
git but this  *will* share your database with others  if you push your
repository, i would not recommend using this method.

cage2:
Hi @gopar!

We just merged some modification to the code that allow to change database annotation during your Emacs session. Even this does not exactly address your request i believe could be a good starting point for other people to implement a mechanism to load different annotations depending from the git branch you are using. In my opinion, this package should not deal with VCS at all because i think expanding the program in this direction will conflict whit the main goal of the package.

But at least there is a starting point! :)

[...]

PR #69, opened 2020-06-03
- added a couple of functions to allow an user to change annotation database

cage2:
Hi Bastian!

With this PR i tried to address (tangentially, to be honest) the problem exposed in #68.

The proposed modifications supposed to reload the database that contains the annotations and then iterate on all the  buffers that uses `annotate-mode` refreshing the annotation in each buffer.

It is a bit crude but my impression is that works.

Honestly i do not think this package should deal with control version system, like asked in #68, l but  i also think that using this function @gopar or other users could be able write some code to get a different annotation file per git branch.

Bye!
C.

### DIFF
--- a/annotate.el
+++ b/annotate.el
@@ -2489,5 +2489,51 @@ annotation, like this:
 
 ;;;; end of filtering: parser, lexer, etc.
 
+;;;; switching database
+
+(defun annotate-buffers-annotate-mode ()
+ "Returns a list of all the buffers that have
+annotate minor mode active"
+  (let ((all-buffers (buffer-list)))
+    (cl-labels ((annotate-mode-p (buffer)
+                  (with-current-buffer buffer
+                    (and (boundp 'annotate-mode)
+                         annotate-mode))))
+      (cl-remove-if-not #'annotate-mode-p all-buffers))))
+
+(cl-defun annotate-switch-db (&optional (force-load nil))
+ "Ask the user for a new annotation database files, load it and
+refresh all the annotations contained in each buffer where
+annotate minor mode is active.
+
+If `force-load' is non nil no prompt asking user for confirmation
+about loading the new file is shown.
+
+Note: this function will attempt to load (compile and
+eval/execute) the content of the file as it was elisp source
+code, always use load files from trusted sources!"
+  (interactive)
+  (let ((new-db (read-file-name "Database file location: ")))
+    (when (not (annotate-string-empty-p new-db))
+      (if (file-exists-p new-db)
+          (let* ((confirm-message "Loading elisp file from untrusted source may results in severe security problems. Load %S? [y/N]")
+                 (load-file-confirmed (if force-load
+                                          t
+                                        (string= (read-from-minibuffer (format confirm-message
+                                                                               new-db))
+                                                 "y"))))
+            (if load-file-confirmed
+                (progn
+                  (setf annotate-file new-db)
+                  (cl-loop for annotated-buffer in (annotate-buffers-annotate-mode) do
+                           (with-current-buffer annotated-buffer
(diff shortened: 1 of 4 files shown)


bastibe:
Looks good to me!

In general, I am a bit reluctant to add code for fringe use cases that increase the maintenance surface of the project. But seeing that the change is contained in two functions and does not change existing code, it seems harmless enough.

Do you see additional value in this code beyond #68? If not, it might be wise to put it into the FAQ instead of annotate's source code. But I'll leave that decision up to you.

cage2:
On Mon, Jun 08, 2020 at 08:21:31AM -0700, Bastian Bechtold wrote:

Hi Bastian!! :)

> Looks good to me!

I have updated the documentation and the other file as usual.

Also i have  added a little patch to remove  the annoying message that
was  printed   about  an  unsaved   modified  buffer  when   the  only
modification was actually just reloading annotate-mode.

if OK i am going to merge this patch in the next hours.

> In general, I am a bit reluctant to add code for fringe use cases that increase the maintenance surface of the project. But seeing that the change is contained in two functions and does not change existing code, it seems harmless enough.

I am surprised too that this  modification took so few changes in the
code to be effective, this is all Emacs merit of course! :)

For bugs blame me instead! :D

>
> Do you see additional value in this code beyond #68? If not, it might be wise to put it into the FAQ instead of annotate's source code. But I'll leave that decision up to you.

I would  never added this feature  if not asked but,  someway, i think
could be useful now.  So i am  happy that we got the issue report #68;
probably this could even become some sort of scaffolding to build some
complicated  kind of  database management  (database of  annotation, i
mean). In  my opinion these  features (interfacing with git  bzr, etc)
should not be  the goal of this library  but i think it is  OK to make
life  easier  for people  that  want  to write  some  code  on top  of
annotate.el.

So i think leaving these functions in  the main code could be a general
improvement for this package.

Bye!!!
C.

annotate.el
```diff
@@ -7,7 +7,7 @@
 ;; Maintainer: Bastian Bechtold
 ;; URL: https://github.com/bastibe/annotate.el
 ;; Created: 2015-06-10
-;; Version: 0.7.0
+;; Version: 0.7.1
```
bastibe: One minor nitpick: In the semantic versioning convention, we should bump the minor version, not the bugfix version. In short, the convention says that major version changes denote breaking changes, minor version changes are feature additions, and bugfix version changes only fix bugs.

Since this is a feature addition, it should result in 0.8.0.

(Although we might want to postpone that a bit if there are more things you'd like to add before pushing a new version)

bastibe:
Thank you for your considerations. This looks great! Apart from the version change, feel free to merge when you're ready.

cage2:
On Tue, Jun 09, 2020 at 11:49:42PM -0700, Bastian Bechtold wrote:

Hi Bastian!

> @bastibe commented on this pull request.
>
>
>
> > @@ -7,7 +7,7 @@
>  ;; Maintainer: Bastian Bechtold
>  ;; URL: https://github.com/bastibe/annotate.el
>  ;; Created: 2015-06-10
> -;; Version: 0.7.0
> +;; Version: 0.7.1
>

> One minor nitpick: In the  semantic versioning convention, we should
> bump the minor version, not the bugfix version.

Right!

>In short, the convention says that major version changes denote breaking changes, minor version changes are feature additions, and bugfix version changes only fix bugs.
>
> Since this is a feature addition, it should result in 0.8.0.

That is absolutely true. My fault, :( I am pushing the fix right know.

Thank  you again  for reviewing  the  code so  carefully, this  always
results in very precious suggestions!

> (Although we might want to postpone that a bit if there are more things you'd like to add before pushing a new version)

I think i am OK, let's see what code can "blossom" from this patch! :)

Bye!
C.

PR #72, opened 2020-06-26
- Added file argument 'database-file-path' to 'annotate-switch-db'.

cage2:
Hi Bastian!

A little change that can help people using the function mentioned in the subject; as the new argument is an optional argument
and appears after the only optional argument in the previous version of the code this will not break the API for other users
(even if i doubt there was any :)).

Also some typos fixed.

Bye!
C.

### DIFF
--- a/Changelog
+++ b/Changelog
@@ -1,3 +1,10 @@
+2020-06-26
+	* annotate.el (annotate-switch-db)
+	- Added file argument 'database-file-path',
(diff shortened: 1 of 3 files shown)


bastibe:
Good idea!

Merge when you're ready.

cage2:
Hello!!
> Good idea!
> 
> Merge when you're ready.

Sure! :)

Bye!
C.

Issue #74, opened 2020-06-30
Buffers marked as modified


articuluxe:
Hello and thank you for the package.  It's neat.  

As soon as I add an annotation, the buffer is marked as modified. I wonder if there is a way to get around this.  The actual buffer contents to be written to disk have not changed; and only the annotation db needs to be updated.  Is there a way to inhibit the modified status?

Thank you.

cage2:
On Tue, Jun 30, 2020 at 08:31:23AM -0700, articuluxe wrote:

Hi articuluxe!

> Hello and thank you for the package.  It's neat.

Thank you! :)

> As soon as I add an annotation,  the buffer is marked as modified. I
> wonder if  there is  a way  to get around  this.  The  actual buffer
> contents  to be  written  to disk  have not  changed;  and only  the
> annotation db  needs to be updated.   Is there a way  to inhibit the
> modified status?

Yes, i think there  is a way to get what you asked  for (but there are
caveat, please read below):

The places where  the code explicitly set the 'modified'  bit "on" are
two:

when all the annotations are removed from buffer:

https://github.com/bastibe/annotate.el/blob/master/annotate.el#L1322

and when a single annotation is created/modified/erased:

https://github.com/bastibe/annotate.el/blob/master/annotate.el#L402

If you  comment out the  form highlighted  you will get  something you
want; but, unfortunately, this is not enough! :)

If you take a look at:

https://github.com/bastibe/annotate.el/blob/master/annotate.el#L1595

in this  local function  (delete) the buffer  is actually  modified by
'annotate--remove-annotation-property'. So  the best way to  deal with
this (probably unintended, implicit) modification is to make something
like what was done here before:

https://github.com/bastibe/annotate.el/blob/master/annotate.el#L1098

That is,  saving the 'modified'  bit before computation  and restoring
after (probably a macro could help here).

I think the latter could be very likely a bug (there is no need to set
modification bit in the local  function as 'annotate-annotate' will do
anyway) thank you for helped me spotting this! But for the first i have
mixed feeling.

If you does not  mark the buffer as modified when  a change -*only* in
the annotations- occurs  the hook 'after-save-hook' will  not be fired
and you will  actually loose the annotations if you  do not explicitly
call 'annotate-save-annotations'  and close  the buffer.  This  is not
something i would like happening to me! ;)

Anyway i have to investigate the bug above for potential regression (i
guess there are not going to be  regressions but some test or at least
[...]

Issue #73, opened 2020-06-30
Info is a dependency


articuluxe:
The usage of Info-current-file means that I had to (require 'info) or encounter an unknown variable error.  I would think this should just be added to the top of annotate.el.

cage2:
On Tue, Jun 30, 2020 at 08:30:01AM -0700, articuluxe wrote:

Hi articuluxe!

> The usage of  Info-current-file means that I had  to (require 'info)
> or encounter an  unknown variable error.  I would  think this should
> just be added to the top of annotate.el.

Wow, it is  actually possible to install Emacs  without info? Genuinely
surprised about that!

I  wonder what  happens if  an user  already have  loaded info  before
loading this package (so, loading info twice).

Now  the  problem  is,  is there a guarantee  that  code  inside  info  was
idempotent?   For example  i think  loading two  times annotate  could
generate some weird behaviour.

Is there a way to conditionally require a package?

I mean: if info is not loaded then (require 'info).

Should not  be user responsibility to  load info (in their  init file)
before loading annotate?

I this  case i think  i should  patch the documentation  to explicitly
state that  info should be  loaded (even  with the 'require'  form you
suggested) to use this package.

I think i need some help here. :)

Bye!
C.

EDIT: added word

bastibe:
I think multiple `require` do not cause problems. It's simply a no-op if it has been loaded elsewhere.

But it's true, if we use `info` functionality, we need to `require` it first.

cage2:
On Wed, Jul 01, 2020 at 12:14:20AM -0700, Bastian Bechtold wrote:

Hello Bastian!

> I think multiple `require` do not cause problems. It's simply a no-op if it has been loaded elsewhere.

Good, my only doubt was if it was harmful to load it multiple times.

>
> But it's true, if we use `info` functionality, we need to `require` it first.

Ok i am going to patch the code with the required "require". ;)

Bye!
C.

cage2:
Hi @articuluxe!

I think the bug you reported is fixed after merging #75 , thank you for reporting! :)

Bye!
C.

articuluxe:
Thank you!

cage2:
On Fri, Jul 03, 2020 at 09:26:33AM -0700, articuluxe wrote:
> Thank you!

You're welcome! :)

C.

PR #75, opened 2020-07-01
Added dependency on info and removed duplicated operation

cage2:
Hi Bastian!

This PR should address #73 and part of #74.

The bug was setting twice the buffer as modified, it was actually harmless but i can see no reason to keeps it anyway! :)

Bye!
C.

### DIFF
--- a/Changelog
+++ b/Changelog
@@ -1,3 +1,10 @@
+2020-07-01
+	* annotate.el (annotate-change-annotation)
+
+	- ensured that the modified status of as buffer is not modified by the local
+	function 'delete' in 'annotate-change-annotation';
+	- added dependency on 'info'.
+
 2020-06-26
 	* annotate.el (annotate-switch-db)
 	- Added file argument 'database-file-path',
--- a/NEWS.org
+++ b/NEWS.org
@@ -120,3 +120,6 @@
 
 - 2020-06-26 V0.8.1 Bastian Bechtold, cage ::
   Added file argument to ~annotate-switch-db~.
+
+- 2020-07-01 V0.8.2 Bastian Bechtold, cage ::
+  Added explicit dependency on 'info' and removed a duplicated operation.
--- a/annotate.el
+++ b/annotate.el
@@ -50,6 +50,9 @@
 ;; can not be annotated.
 
 ;;; Code:
+
+(require 'info)
+
 (require 'cl-lib)
 
 ;;;###autoload
--- a/annotate.el
+++ b/annotate.el
@@ -58,7 +58,7 @@
 ;;;###autoload
 (defgroup annotate nil
   "Annotate files without changing them."
-  :version "0.8.1"
+  :version "0.8.2"
   :group 'text)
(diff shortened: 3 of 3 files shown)


bastibe:
I don't think I understand the bug. It seems to me that the proposed change would leave the buffer-modified-bit unchanged when an annotation is deleted, which doesn't seem to make sense. Could you explain the reasoning behind that change?

Otherwise, this looks good.

cage2:
On Thu, Jul 02, 2020 at 04:47:48AM -0700, Bastian Bechtold wrote:

Hello Bastian!

> I don't think I understand the bug. It seems to me that the proposed change would leave the buffer-modified-bit unchanged when an annotation is deleted, which doesn't seem to make sense. Could you explain the reasoning behind that change?

Sure! I  think explaining my change  can also help me  realize if this
patch was a good idea or not! :)

I noticed that in the code we explicitly set the buffer as modified in
just two points:

when all the annotations are removed from buffer:

https://github.com/bastibe/annotate.el/blob/master/annotate.el#L1322

and when a single annotation is created/modified/erased:

https://github.com/bastibe/annotate.el/blob/master/annotate.el#L402

But, for  example, the code does  not set the buffer  as modified when
load annotation  from the database, in  this case it just  restore the
value of the modified status before the loading of annotations occurs.

This behaviour does makes sense to  me (explicit marks the buffer only
when specific changes occurs regarding  the annotations) and is what i
would expect from the package.

But there  is a  single (as far  as i know)  point where  the modified
status is set to true in an  implicit way, when a single annotation is
deleted.

https://github.com/bastibe/annotate.el/blob/master/annotate.el#L1595

This  introduce an  implicit side  effect that  could be  difficult to
track   when  a   function  calls   'annotate-change-annotation'  and,
moreover, the only function that calls 'annotate-change-annotation' is
'annotate-annotate'  where  the buffer's  status  is  set as  modified
anyway (at the end of the function's body).

I     think    removing     the     side     effect    hidden     into
'annotate-change-annotation' effect  from 'annotate-change-annotation'
could  prevents   some  bugs   in  the  future   where  (and   if  ;-)
'annotate-change-annotation' is called by some other functions.

Moreover we  could even  implement what articuluxe  asked in  #74, for
example using a customizable variable  that prevent marking the buffer
as modified.

Looking forward for your opinion and, if i was not able to explain my
idea feel free to ask, of course! :)

Bye!
C.

bastibe:
I see! Thank you for the explanation, I had missed the fact that `annotate-annotate` already marked the buffer as modified. It all makes sense now, and I agree with your assessment. Good catch on that bug!

cage2:
Hi Bastian!

> I see! Thank you for the explanation, I had missed the fact that `annotate-annotate` already marked the buffer as modified. It all makes sense now, and I agree with your assessment. Good catch on that bug!

I have to thanks @articuluxe because otherwise i think i never spotted it :)

OK, so going to merge in a second

Bye!
C.

Issue #80, opened 2020-09-15
Working with visual-line-mode


oatmealm:
Hi, thanks for a wonderful mode!

Wanted to ask if annotate-mode is designed to work with `visual-line-mode` for example in `org-mode` buffer?

I have it set to 'margin, but the effect is new line, with text aligne to the left. 

![gnome-shell-screenshot-VWP0Q0](https://user-images.githubusercontent.com/68159077/93204609-62ff4000-f724-11ea-845e-0b31b9ed55ce.png)

cage2:
On Tue, Sep 15, 2020 at 04:24:21AM -0700, oatmealm wrote:
> Hi, thanks for a wonderful mode!

Hi!

Always happy if this library is useful for other people! :)

> Wanted to ask if annotate-mode is designed to work with `visual-line-mode` for example in `org-mode` buffer?
>
> I have it set to 'margin, but the effect is new line, with text aligne to the left.
>

To  be  honest   i  do  not  think  there   was  something  preventing
annotate-mode  to works  fine with  visual-line-mode.  I  can see  two
problem here: the wrong text flushing  in the annotation (but i wonder
if is possible to mix two language with different line flushing in the
same  buffer, actually  i have  no idea  but if  emacs could)  and the
newline.

Please  can you  share  a  little org  file  thah  shown the  problem?
Because  i  tried  to  reproduce  the second  bug  (newline)  with  no
luck. Also  in the  image you provided  how much is  the width  of the
window  containing the  buffer?   Have  you tried  to  enlarge it  8if
possible of course) and see what happens?

The problem  with text flushing  is simple to  reproduce but i  do not
know  how much  this  package can  do  to prevent  that  (any hint  is
welcome!).

Bye!
C.

oatmealm:
Hi @cage2 ! Thanks for taking the time to have a look. Attaching and org file with the annotations.
[demo.zip](https://github.com/bastibe/annotate.el/files/5226948/demo.zip)

Notes:

1. The buffer in the screenshot has writteroom-mode, visual-line-mode and variable-pitch mode enabled.
2. The variable `bidi-parapgraph-direction` is set to 'nil, so flushing/alignment of English/Arabic/Hebrew is automatic.

I don't know what controls the comment's appearance... is it affected by any of the modes enabled in the buffer? 

The comment in English seems to be flashed to the left, but the rest of the text is wrong.

The second screenshot is with writeroom-mode disabled.

[...]

Issue #79, opened 2020-09-06
Query Language link incorrect


andrelkin:
In the description `Query Language` is referred with an incorrect link, the correct one is https://github.com/bastibe/annotate.el#query-language.
Could someone please fix that?

cage2:
Hi @andrelkin !

Thank you for your report, never noticed this issue! Seems to me that the link in the org-mode file was correct and perhaps the problem could be in the rendering engine of github. I just checked the source file and the generated HTML file from Emacs and in both formats the link works.

Any idea how can i forward this to github folks?

Bye!
C.

cage2:
Hi!

I have contacted github about this, in the meanwhile maybe would be a good idea to remove the link from the README.

Bye!
C.

andrelkin:
To remove looks good. Thank you for addressing it! I am really looking forward to make use
of your tool at my work place (and who knows if that won't be followed with contributions to the project :-)!).

cage2:
Hi @andrelkin !

So the folks at github kindly answered my question and wrote tho change the internal link:

```
[[Query language]]
```

with 

```
[[https://github.com/bastibe/annotate.el#query-language][Query Language]]
```
This is exactly your first suggestion (and form their -i mean github- point of view a reasonable solution). :-D

But, instead, i feel this is a workaround that will break the ORG file so I am inclined to remove the link anyway. But i am happy to discuss using the first solution (changing, not removing, the link).

I think the right thing to do is to ask @bastibe (the maintainer)  for his opinion.

Anyway i hope that this package will be useful for your work (or fun :-) ) and i will be happy if you decided to contribute to this package! :)

The code, i think, is not too difficult to grasp and there are docstrings for each function and variable. There are some parts a bit convoluted so feel free to ask me if you need explaination, i'll try my best to answer your questions! :)

Bye!
C.

bastibe:
I completely agree with you, @cage2, on all points. Github's org parser is broken, and I'd rather have no link than a link that works on Github, but not in Emacs. And annotate.el is a cool code base indeed, in no small part thanks to @cage2's contributions!

cage2:
Hello everyone!

> I completely agree with you, @cage2, on all points. Github's org parser is broken, 

yes it is, i think they made a lot of improvements lately but i feel someway org is still a second class citizen here.

> and I'd rather have no link than a link that works on Github, but not in Emacs. 

OK! @andrelkin  if you agree i would remove the internal link.

[...]

PR #81, opened 2020-09-20
Popup annotation

cage2:
Hi!!

This is a draft of what suggested by @oatmealm in #80.

I used the `help-echo` property as discussed in the same thread.

Honestly i do not like the aesthetic of the popup on my system but this is totally subjective, of course. And, moreover, i agree this feature can  help some users. So OK, i will deal with it! :-D

To activate the popup system a new customized variable has been added: `annotate-use-echo-area`.

I wonder what i am missing because this was too simple. ;-)

Bye!
C.

edit: also i removed the internal link in the README as pointed out in #79

### DIFF
--- a/Changelog
+++ b/Changelog
@@ -1,3 +1,31 @@
+2020-09-29
+        * README.org, annotate.el
+	- updated README;
+	- increased version;
+	- updated NEWS and Changelog.
+
+2020-09-22
+        * annotate.el
+	- added   command   'annotate-summary-of-file-from-current-pos';
+	this command shows a summary window that contains the annotation
+	in the  active buffer  that appears after  the cursor  position;
+	- added docstrings.
+
+2020-09-20
+        * annotate.el
+	- fixed indentation.
+
+2020-09-20
+        * annotate.el
+	- Using "help-echo" to print annotations instead of render it on the
+        buffer.
+	This feature is optional and can be activated using the
+        customizable variable 'annotate-use-echo-area'
+
+2020-09-20
+        * README.org
+	- Removed internal link because of issue #79.
+
 2020-08-11
 	* annotate.el (annotate-annotate, annotate-load-annotation-data
 	               annotate-create-annotation, annotate-summary-query-parse-note,
--- a/NEWS.org
+++ b/NEWS.org
@@ -126,3 +126,7 @@
 
 - 2020-08-11 V0.8.3 Bastian Bechtold, cage ::
   Some function now signal errors where appropriate.
+
+- 2020-09-29 V0.9.0 Bastian Bechtold, cage ::
+  Added two new styles to render the annotation: using "pop-up" style
+  or via a specializated summary window.
--- a/README.org
+++ b/README.org
@@ -78,7 +78,8 @@ can take advantage of its packages generated files management.
     A window with a list of annotated files together with their
     annotations is shown. If ~annotate-summary-ask-query~ is non nil
     (default is ~t~) then a prompt is shown where the user can insert
-    a query to filter the annotation database, see [[Query Language]].
+    a query to filter the annotation database, see "Query Language"
+    below.
 
     The summary window allow editing and removing of annotation using
     the provided buttons.
--- a/README.org
+++ b/README.org
@@ -109,6 +109,46 @@ as comments into the current buffer, like this:
      - ~annotate-integrate-highlight~
      - ~annotate-fallback-comment~
 
+* Alternative visualization of annotations
+
+For typographically difficult scenarios (or just because you prefer
+it), such as variable-width fonts or overlay-heavy modes, the default
+visualization system that renders the annotation into the buffer could
+not properly works.
+
+In this case the users can switch to a "pop-up" style annotation
+setting to a non-nil value the variable ~annotate-use-echo-area~.
+
+When such variable's value is not null, moving the mouse pointer over
+the annotated text will temporary show the annotation.
+
+The actual visuals of this "pop-up" can be different depending of your
+system's setup (see
+[[https://github.com/bastibe/annotate.el/pull/81][this pull request]]
+for a couple of examples.
+
+Another alternative way to show annotations is provided by the command:
+~annotate-summary-of-file-from-current-pos~.
+
+Calling this command will show a summary window that prints all the
+annotations related to annotated text that appears (in the active
+buffer) beyond the current cursor position.
+
+**** related customizable variable
+     - ~annotate-use-echo-area~
+
+* Other commands
+
+** annotate-switch-db
+
+This command will ask the user for a new annotation database files,
+load it and refresh all the annotations contained in each buffer where
+annotate minor mode is active.
+
+See the docstring for more information and
+[[https://github.com/bastibe/annotate.el/issues/68][this thread]]
+for a possible workflow where this command could be useful.
+
 * More documentation
 
  Please check ~M-x customize-group RET annotate~ as there is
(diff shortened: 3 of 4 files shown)


oatmealm:
Will have a look shortly... sounds helpful. 

I was just thinking, what if the annotations were shown in a side bar I.e a separate window that can be toggled on and off, while notes are synced with the main conetent buffer. 

org-noter has this nice semi-automatic (I think you need to 'C-M .' first) sync facility that advances or backtracks the notes buffer in respect to your position in the pdf-view buffer.

cage2:
On Sun, Sep 20, 2020 at 08:24:03AM -0700, oatmealm wrote:

Hi!

> Will have a look shortly... sounds helpful.

Good! :)

> I was just thinking, what if the annotations were shown in a side bar I.e a separate window that can be toggled in and out while notes are synced with the cone tent buffer.
>
> org-noter has this nice semi-automatic (I think you need to 'C-M .' first) sync facility that advances or backtracks the notes buffer in respect to your position in the pdf-view buffer.

You  could  (sort  of)  use  the summary  window  for  something  like
that. Just search the database for the current file that the buffer is
visiting.

This will show all the annotation contained in the buffer, though.

Although maybe  could not be  too difficult  to write a  function that
makes that search for the user and  cuts the summary to shown only the
annotations that annotates text beyond  the current cursor position in
the buffer.

Interesting! But I have to think a bit about that! 🤔

Bye!
C.

oatmealm:
Looks nice :+1: 

In the echo area "הערה מס 1" (comment no 1) ...

BTW the pointer was on the annotation in the buffer, but wasn't captured... the selection in blue is there by accident.

Must say though it's still probably more useful when in the buffer closer to the text. 

gnome-shell-screenshot-5KLIR0

cage2:
Hi!

> Looks nice +1

That's good! :)

> In the echo area "הערה מס 1" (comment no 1) ...

I think this is dependents from the Emacs configuration (perhaps at compile time)   or graphical toolkit used, please take a look at the results i get in the image below:

https://www.autistici.org/interzona/img/misc/popup.png

Yes seems there is an broken icon, no idea why! ;-)

> Must say though it's still probably more useful when in the buffer closer to the text.

I agree, i think the old behaviour could be the default.

Anyway i like the idea to use a window to summarize the annotation in a single file! :)

Bye!

cage2:
Hi @oatmealm!

I have added a new command: `annotate-summary-of-file-from-current-pos` that shows annotation in the active buffer that appears after the cursor position. Can you, please, take a look if this feature is more or less the suggestion you made in your first comment of this thread?
Thank you!

C.

bastibe (approved):

bastibe:
This looks like a useful addition to annotate.el, in particular for typographically difficult scenarios such as variable-width fonts or overlay-heavy modes. I like it. Good work!

oatmealm:
> Can you, please, take a look if this feature is more or less the suggestion you made in your first comment of this thread?


Sorry. Didn't see this before. Will have a look!

cage2:
On Mon, Sep 28, 2020 at 03:28:04AM -0700, Bastian Bechtold wrote:

Hi Bastian!

> This looks like a useful  addition to annotate.el, in particular for
> typographically difficult scenarios such  as variable-width fonts or
> overlay-heavy modes. I like it. Good work!

Thank you!   Part of the merits  goes to oatmealm that  suggested both
and is helping with testing! Honestly  this was not too difficult as i
am building on  top of existing package's code (and  Emacs of course).
Things seemes to go quite  smooth...until fonts enter in the equation!
😉😉

Seriously, i  think we  are giving  users more  options to  adapt this
package  in their  setup.  But  i would  still  like  to explore  your
proposal about using Unicode space (thin, hair and similar).

Now, may i ask you two question:

- i think  the content  of your  message, slightly  rephrased perhaps,
  could go  into the readme  as this  provide a possible  solution for
  persons  that uses  variable-width fonts,  what do  you think  about
  that?

- which one supposed to be the next version number (after merging this
  PR): 0.8.4 seems to low but 0.9.0  to much instead, wich one is your
  opinion?

Bye!
C.

cage2:
On Mon, Sep 28, 2020 at 05:10:07AM -0700, oatmealm wrote:

Hi oatmealm!

> > Can you, please, take a look if this feature is more or less the suggestion you made in your first comment of this thread?
>
>
> Sorry. Didn't see this before. Will have a look!

No need to be sorry, i am not in a hurry ;), just take your time :)

Bye!
C.

bastibe:
> - i think the content of your message, slightly rephrased perhaps, could go into the readme as this provide a possible solution for persons that uses variable-width fonts, what do you think about that? 

Good idea!

> - which one supposed to be the next version number (after merging this PR): 0.8.4 seems to low but 0.9.0 to much instead, which one is your opinion?

Since this is a feature addition, it should go into the minor version, i.e. 0.9.0. I generally try to adhere to the semantic versioning convention of *major.minor.bugfix*, where major versions are for big milestones or backwards-incompatible changes, minor versions are for additions, and bugfix versions are for bugfixes only.

oatmealm:
@cage2 ~~What should I check out? Couldn't find this function on you pull request #81~~ ...

Ok, my bad. Problems is that on this branch I'm having serious problems with showing annotations. They sometimes appear, sometime not, and never written to the db...

I keep getting "The annotation database is empty" when I try the new command, and indeed the file pointed to by 'annotate-file' never gets written to...

cage2:
Hi @oatmealm 

> @cage2 ~What should I check out? Couldn't find this function on you pull request #81~ ...
> 
> Ok, my bad. Problems is that on this branch I'm having serious problems with showing annotations. They sometimes appear, sometime not, and never written to the db...
> 
> I keep getting "The annotation database is empty" when I try the new command, and indeed the file pointed to by 'annotate-file' never gets written to...

I am sorry that you are having trouble, can you share a little file where the problem shows as you did before?

Bye!
C.

cage2:
Hi @bastibe !

> > * i think the content of your message, slightly rephrased perhaps, could go into the readme as this provide a possible solution for persons that uses variable-width fonts, what do you think about that?
> 
> Good idea!

Very well!

> > * which one supposed to be the next version number (after merging this PR): 0.8.4 seems to low but 0.9.0 to much instead, which one is your opinion?
> 
> Since this is a feature addition, it should go into the minor version, i.e. 0.9.0. I generally try to adhere to the semantic versioning convention of _major.minor.bugfix_, where major versions are for big milestones or backwards-incompatible changes, minor versions are for additions, and bugfix versions are for bugfixes only.

I think you already explained me the convention you adopted for this program but still semantic versioning just do not enter in my mind. And i even think it is a good convention! :-D

Anyway i can see now that 0.9.0  is the right version number, and i promise this is the last time you need to explain semantic version! :-D :-D

Bye!
C.

bastibe:
> Anyway i can see now that 0.9.0 is the right version number, and i promise this is the last time you need to explain semantic version! :-D :-D

No problem at all ;-)

cage2:
Hi @oatmealm!

Sorry to bother you but i wonder if you could provide a way i can reproduce the bug you met with this branch.

Bye!
C.

cage2:
Hi Bastibe!

I can not reproduce  the bug pointed out by oatmealm,  so i propose to
merge it anyway. What is your opinion?

Bye!
C.

bastibe:
> I can not reproduce the bug pointed out by oatmealm, so i propose to merge it anyway. What is your opinion?

If it looks good on your end, I'd say we merge it now, and deal with potential regressions later.

cage2:
Hi @bastibe!

> If it looks good on your end, I'd say we merge it now, and deal with potential regressions later.

Let's merge then! :-)

Bye!
C.

Issue #83, opened 2020-10-25
Automatically expand overlapping annotation


xeruf:
I accidentally placed my annotation on only a few words, but I intended to annotate the whole line:

![image](https://user-images.githubusercontent.com/13354331/97117676-9d85d080-1705-11eb-9f9e-88f5bb57ab3f.png)

When selecting the whole line and pressing `C-c C-a`, I get the following error:

![image](https://user-images.githubusercontent.com/13354331/97117665-7d561180-1705-11eb-92d7-fe5ee4d07021.png)

I would have expected it to widen the annotation to my selection and then simply let me edit the text.
It seems with the current behavior I have to copy the annotation text, delete the annotation and then create it again...

bastibe:
Great idea, actually, to widen the annotation instead of throwing an error.

cage2:
On Mon, Oct 26, 2020 at 01:54:29AM -0700, Bastian Bechtold wrote:

Hi Bastian!

> Great idea, actually, to widen the annotation instead of throwing an error.

I agree that this is a good idea and that i would like to try to implement! :)

Bye!
C.

cage2:
Hi @xerus2000 !

I have just filed a PR  (#87)  that should address your issue, i wonder if i has been able to deal with all the cases when an annotation can overwrite others. Would be great if you could check this branch and makes some minimal testing. Could you?

Bye and thanks in advance!
C.

xeruf:
I'll see, currently busy with other stuff, and annotate is kinda not serving me that well since I often edit files after annotating them ;)

cage2:
On Wed, Dec 02, 2020 at 12:09:23PM -0800, Janek wrote:

Hi!

> I'll see, currently busy with other stuff, and annotate is kinda not
> serving me that well since I  often edit files after annotating them
> ;)

No problem! take your time and thank you anyway! :)

Bye!
C.

PR #84, opened 2020-11-06
Fixed a regression  and some bugs related to  incorrect calculation of multiline annotations.

cage2:
Hi @bastibe !

I introduced some regression and i failed to manage correctly the multi line annotations. :(
The bugs are too difficult for me to explain so i preferred to add use cases to reproduce the issues and proposed fix. Hope this is OK.

There are some thing still missing in this PR (probably just a couple of docstrings) but the bulk of the fix is here, i submitted this PR now so that maybe some user can test this branch and help with testing and i am going to check later what is missing.

Fortunately seems to me that some of the code here should be useful for #83!

Bye!!
C.

``` text
To reproduce the bugs:

legend:

a = annotated text
* = non annotated text

- First bug

Create a multiline annotation using region.

aaaa
aaaa
aaaa    ####

Place the cursor as below.

aaaa
 ^ cursor
aaaa
aaaa    ####

type a character

a****
aaaa
aaaa    ####

The annotated text has a "gap"

Fix proposed: revert to the old (correct behaviour)

Second bug

aaaa
aaaa
aaaa    ####

Place the cursor as below.

aaaa
^ cursor on the first column
aaaa
aaaa    ####

type some text

***
aaa
aaa    ####

Save (C-x C-s)

you  get an  error  on  the echo  area:  "let*:  Wrong type  argument:
overlayp, nil" and the annotations are not correctly saved.

Fix proposed: remove the offending code.

Third bug

a multiline bug as before

aaaa
aaaa
aaaa    ####

place the cursor here:

aaaa
aaaa
^ cursor
aaaa    ####

type some text

aaaa
*****
aaaa    ####

Then annotate the same line (C-c C-a):

aaaa
aaaa    ####
aaaa    ####

we  introduced  a  annotation  in  the gap  of  the  already  existing
multiline annotation.

Fix proposed: prevents annotating text inside an annotation.

```

### DIFF
--- a/Changelog
+++ b/Changelog
@@ -1,3 +1,9 @@
+2020-11-22  cage
+
+        * annotate.el:
+
+	- added more docstrings.
+
 2020-11-12  cage
 
         * annotate.el:
--- a/Changelog
+++ b/Changelog
@@ -1,3 +1,117 @@
+2020-11-12  cage
+
+        * annotate.el:
+
+	- prevented getting shared value for 'annotate-file' in
+        '%load-annotation-data'.
+
+	This  way  we can  ensure  that  'annotate-file' can  be  declared
+	buffer-local and  different annotation  databases can be  used from
+	different buffers.
+
+2020-11-11  cage
+
+        * annotate.el:
+	- fitted 'annotate-position-inside-chain-p' into   'annotate--position-inside-annotated-text-p';
+	- fixed some typos.
+
+2020-11-06  cage
+
+        * annotate.el:
+
+	- Fixed a regression  and some bugs related to
+        incorrect calculation of of multiline annotations.
+
+	To reproduce the bugs:
+
+	legend:
+
+	a = annotated text
+	* = non annotated text
+
+	- First bug
+
+	Create a multiline annotation using region.
+
+	aaaa
+	aaaa
+	aaaa    ####
+
+	Place the cursor as below.
+
+	aaaa
+	^ cursor
+	aaaa
+	aaaa    ####
+
+	type a character
+
+	a****
+	aaaa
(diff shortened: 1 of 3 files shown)


bastibe:
I currently don't have the time to truly look into this, but I trust you completely to do the right thing!

cage2:
On Mon, Nov 09, 2020 at 01:51:04AM -0800, Bastian Bechtold wrote:

Hi Bastian!

> I currently don't have the time to truly look into this,

No problem! Just hope everything is fine.

> but I trust you completely to do the right thing!

Thank you! I promise i will do my best! :)

Bye!
C.

bastibe:
Hi Cage,

> > I currently don't have the time to truly look into this,

> No problem! Just hope everything is fine.

We are doing very well, thank you. But we had a new baby a few weeks ago, which leave little room for side projects at the moment. I hope you are doing well also!

cage2:
On Wed, Nov 11, 2020 at 12:01:15AM -0800, Bastian Bechtold wrote:
> Hi Cage,

Hi!!

> > > I currently don't have the time to truly look into this,
>
> > No problem! Just hope everything is fine.
>
> We are doing very well, thank you. But we had a new baby a few weeks ago, which leave little room for side projects at the moment.

My congratulations! So you are starting a new adventure! ;-)

> I hope you are doing well also!

I am  quite OK,  things are  getting complicated  lately here  where i
live, because of the pandemic, but i  am still alive in good shape and
confident that everything is going to be alright soon! :)

Bye!
C.

cage2:
Hi!

I think this PR is ready! I am going to merge it in the next hours.

Bye!
C.

PR #87, opened 2020-12-01
two more features and a bugfix.

### LINKED ISSUE #83: Automatically expand overlapping annotation

### LINKED ISSUE #85: Add a customizable variable to suppress file modification warnings
The package keeps a persistent hash of an annotated file to check if it was modified; if there is a mismatch between the hash in the database and the one calculated from the actual file a warning is printed.

A customizable variable could be added to suppress the printing of the warning messages.

Bye!
C.

cage2:
Hi!

- overwriting of annotations.

That is,  user can place an  annotation on top of  an already existing
one. The new will delete overlapped portion of the old annotation.

This   feature   should   not    allow   to   break   an   annotation,
though. Annotations can not overlaps.

- added a new customizable variable: 'annotate-warn-if-hash-mismatch'

  when  nil prevent  printing of  warning when  annotation database's'
  hash and file has do not match.

- fixed bug in alternating coloring of annotation and underlined text;

- updated README;

- fixed some typos.

This patch, if correct, should fix #85 and also fix #83.

Bye!

### DIFF
--- a/Changelog
+++ b/Changelog
@@ -1,3 +1,38 @@
+2020-12-16 cage
+
+	* Changelog, NEWS.org, annotate.el:
+
+	Updated version and documentations.
+
+2020-12-07  cage
+
+        * README.org, annotate.el:
+
+	- fixed more typos.
+
+2020-12-01  cage
+
+        * README.org, annotate.el:
+
+	- allow overwriting (even partial)  of annotations.  That is, user
+	can place an annotation on top of an already existing one. The new
+	will  delete  overlapped  portion  of the  old  annotation.   This
+	feature    should   not    allow   to    break   an    annotation,
+	though. Annotations can not overlaps.
+
+	- added a new customizable variable:
+
+        'annotate-warn-if-hash-mismatch'  when  nil  prevent  printing  of
+	warning  when annotation  database's'  hash and  file  has do  not
+	match;
+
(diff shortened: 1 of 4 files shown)


bastibe:
Hi! Looks great! I can see the multiple annotation databases being quite useful!

cage2:
Hi @bastibe !

Yes this is a good feature indeed,  acknowledgment to @gopar for proposing and testing the whole machinery! :-)

Bye!
C.

cage2:
Hi @bastibe !

I think this can be merged and fixed, if needed, later.
What do you think?

Also is my impression that this could be the 1.0 release! :) :)  Right?

Bye!
C.

bastibe:
Hey, sounds good to me! 1.0 for Christmas!

cage2:
Hi bastibe!

> Hey, sounds good to me! 1.0 for Christmas!

That's great! 🎊 🎉

Just let me  update the version number,  and i am going  to merge this
PR!

Bye!
C.

PR #93, opened 2021-01-06
prevented side effects of 'split-string'

cage2:
Hi @bastibe ! 

I  think  i  have added  this  bug  many  months  ago and  was  hiding
unnoticed! :) It  was discovered by chance only because  of the latest
code changes related to multi-line annotated text.

Truly i wonder how did the code worked until now! :astonished: 

Bye!
C.

### DIFF
--- a/Changelog
+++ b/Changelog
@@ -1,3 +1,10 @@
+2021-01-06  cage
+
+        * annotate.el:
+
+	- prevented side effects of 'split-string'.
+
+
 2020-12-23  cage
 
         * annotate.el:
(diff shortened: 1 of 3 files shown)


bastibe:
> Truly i wonder how did the code worked until now! 😲

If only I knew that about my code ;-)

cage2:
On Fri, Jan 08, 2021 at 12:26:12AM -0800, Bastian Bechtold wrote:
> > Truly i wonder how did the code worked until now! 😲
>
> If only I knew that about my code ;-)

This  two  sentences seems  to  explain  why  programming is  a  weird
activity! :-D

Bye!
C.

Issue #90, opened 2020-12-16
render overlay as a whole block


randomwangran:
In a multi-lines note, there will be some blank lines. These
lines will not be render. This will make the notes a little bit
strange looking.

I have post two images to show what I mean.

The first one is the current state. The second one is the desired
state, which I use GIMP to render.

![image](https://user-images.githubusercontent.com/8197831/102387202-16830500-3f9e-11eb-8eed-a09e6de5fe4d.png)

![2020-12-16_124247](https://user-images.githubusercontent.com/8197831/102387505-75487e80-3f9e-11eb-9476-ce6a4f8f5ba6.png)

cage2:
Hi @randomwangran !

A little question: you added manually (i mean with ^Q ^J two times) the linebreak in the annotation text, right?

Bye!
C.

randomwangran:
I use this function to create new line: `default-indent-new-line`

```
M-j runs the command default-indent-new-line (found in global-map),
which is an interactive compiled Lisp function in ‘simple.el’.

It is bound to C-M-j, M-j.

(default-indent-new-line &optional SOFT)

Break line at point and indent.
If a comment syntax is defined, call ‘comment-line-break-function’.

The inserted newline is marked hard if variable ‘use-hard-newlines’ is true,
unless optional argument SOFT is non-nil.
```

Let me know if I am make it clear. Thank you.

cage2:
On Thu, Dec 17, 2020 at 10:11:56AM -0800, randomwangran wrote:

Hi randomwangran!

I just pushed some modification to add this feature on the package.

May i ask to do some testing if it works fine or not?
Thanks in advance!
C.

NB: the rendering of the "box"  around the multiline annotated text is
slightly different  from the mock up you  posted as i do  not extend the
background  (i mean,  non printable  text with  "coral" as  background
color) to the border of the window (this could results in rendering if
the window  is shrunk even  by single  column), but just  enough to
wrap the text.

randomwangran:
Hi @cage2 

c103381862f5124e5871c37282e67f3d2e0c588a works perfectly on my side!

> the rendering of the "box"  around the multiline annotated text is
slightly different 

Your version is the best. I mocked up just for illustration purpose. No worried about it.

Thanks for your effort!

cage2:
On Sun, Dec 20, 2020 at 08:18:17AM -0800, randomwangran wrote:
> Hi @cage2

Hi!

[...]

PR #95, opened 2021-01-14
Improving visuals of margin notes

cage2:
Hi @bastibe !

With this patch i tried to improve the  appeareance of the notes placed on the border of the window in the same way @randomwangran suggested for annotation placed on theyr own line (see https://github.com/bastibe/annotate.el/issues/90).

The code was mostly there fortunately! :)

Bye!
C.

### DIFF
--- a/Changelog
+++ b/Changelog
@@ -1,3 +1,9 @@
+2021-01-14 cage
+
+        * annotate.el:
+
+	- added padding also for notes placed on the margin of the window.
+
 2021-01-06  cage
 
         * annotate.el:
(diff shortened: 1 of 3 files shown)


bastibe:
One step at a time, we fix the bugs. Thank you!

cage2:
On Fri, Jan 15, 2021 at 04:29:19AM -0800, Bastian Bechtold wrote:
> One step at a time, we fix the bugs. Thank you!

My pleasure Bastian! This tool is part of my regular workflow and i am
so happy  that there are  some people that  are using this  package in
ways  i  was not  able  to  predict  (for  example with  an  encrypted
annotated database).

Bye!
C.

Issue #94, opened 2021-01-14
Encrypted annotations


eve-nisse:
Encrypting annotations file requires frequent decryption (and key unlocking) despite annotations.gpg being kept open and decrypted at startup.

Expected behavior: Prefer decrypted annotations file buffer over disk when saving annotations or initializing the mode

cage2:
Hi!

I never thought about annotate encrypted files, interesting!

For key unlocking i wonder if could be useful increase the cache of gpg like explained here:
https://superuser.com/a/624488

I think even epa mode can do something similar (M-x customize-group epa) but i never tried.

If you annotate the file the buffer is marked as modified so epa will try to encrypt anyway.

Probably i could add a customizable variable to prevent buffer to be marked as modified as i proposed here:

https://github.com/bastibe/annotate.el/issues/74

What is your opinion about that proposal? Would this fix your issue?

Bye!
C.

eve-nisse:
Increasing cache time decreases security a bit, so I avoid it and have considered not caching at all. It's a sort of last resort option.

Below is the log for when an annotated 'testing.org' file is saved. My hunch is that being aware of the already open decrypted "annotations.gpg" buffer in "annotate-load-annotations" which is called by "annotate-save-annotations" will fix it. Then only encryption will be needed to save. That's the extent of my elisp ability at the moment.

```
Saving file /home/user/.emacs.d/testing.org...
Wrote /home/user/.emacs.d/testing.org
Decrypting /home/user/.emacs.d/annotations.gpg...done
Encrypting /home/user/.emacs.d/annotations.gpg... [2 times]
Annotations saved.
Reverting buffer ‘annotations.gpg’.
Decrypting /home/user/.emacs.d/annotations.gpg...done
```

cage2:
On Thu, Jan 14, 2021 at 08:12:47AM -0800, EverardNisse wrote:

Hi!

> Increasing cache  time decreases security a  bit, so I avoid  it and
> have  considered not  caching at  all. It's  a sort  of last  resort
> option.
> Below  is  the log  for  when  an  annotated 'testing.org'  file  is
> saved. My  hunch is that being  aware of the already  open decrypted
> "annotations.gpg"  buffer  in "annotate-load-annotations"  which  is
> called  by  "annotate-save-annotations"  will   fix  it.  Then  only
> encryption will  be needed to  save. That's  the extent of  my elisp
> ability at the moment.

[...]

I think i am starting to understand. Probably the bottleneck is in the
[...]

PR #96, opened 2021-01-17
Added  early checking encrypted (GPG) file format

cage2:
Hi @bastibe 

Inspired by the issue reported in  #94 i added the magic number check for GPG encrypted files, from the Changelog:

``` text
    In  many parts  of  the code  we  use 'annotate-guess-file-format',  a
    function that  try to guess the  format of the file  that produced the
    contents of a  buffer. Sometimes (for example buffers  that shows info
    nodes)  this check  is very  expensive because  involves reading  (and
    possibly decompress or even decrypt) the file contents and check for a
    regular expression.
    
    If we could  know in advance that the file  is symmetrically encrypted
    with GPG we could skip other,  more expensive checks for guessing file
    format (like info files).
```
I know that this is probably sort of a corner case (people who annotates encrypted files) but i found no reason to implement it if this can makes life a bit easier for someone :-) (and the code is trivial by the way ;-)).

Bye!
C.

### DIFF
--- a/Changelog
+++ b/Changelog
@@ -1,3 +1,22 @@
+2021-01-15  cage
+
+        * annotate.el:
+
+	- added checking encrypted (GPG) file format.
+
+	In many parts  of the code we  use 'annotate-guess-file-format', a
+	function that  try to guess the  format of the file  that produced
+	the  contents of  a buffer.
+
+	Sometimes (for example  buffers that shows info  nodes) this check
+	is  very   expensive  because   involves  reading   (and  possibly
+	decompress  or even  decrypt) the  file contents  and check  for a
+	regular expression.
+
+	If  we  could know  in  advance  that  the file  is  symmetrically
+	encrypted with GPG we could  skip other, more expensive checks for
+	guessing file format (like info files).
+
 2021-01-14 cage
 
         * annotate.el:
--- a/NEWS.org
+++ b/NEWS.org
@@ -187,3 +187,8 @@
 
   This  version improves  visual  of multilined  notes  placed on  the
   window margins.
+
+- 2021-01-06 V1.1.3 Bastian Bechtold, cage ::
+
+  Optimized the code to speedup  reading and saving of encrypted (with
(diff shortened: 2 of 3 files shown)


eve-nisse:
While my use case is only asymmetric encryption (should've probably mentioned this), I tried to cherry-pick the commit and symmetrically encrypt 'annotations.gpg'. Saving still seems to resolve in decryption of the file even though it's already open as a buffer, much like before. It's possible my test was flawed as it wasn't isolated from my usual easypg setup, so trust your machine first. 

Good work nonetheless.

cage2:
On Sun, Jan 17, 2021 at 02:42:32PM -0800, EverardNisse wrote:

Hi!

> While my use case is  only asymmetric encryption (should've probably
> mentioned this), I tried to cherry-pick the commit and symmetrically
> encrypt 'annotations.gpg'.

Thank you for reviewing this PR.

> Saving still seems to resolve in  decryption of the file even though
> it's already  open as a buffer,  much like before. It's  possible my
> test was flawed as it wasn't isolated from my usual easypg setup, so
> trust your machine first.

I think  your test is  fine as although this  PR was inspired  by your
issue  report  is  actually  meant  for the  opposite  of  your  case:
annotating encrypted files, not keeping annotations in an an encrypted
database.

Sorry for this misunderstanding, but thanks anyway for testing!

>
> Good work nonetheless.

Thank you!

The issue  you reported could be  (sort of) fixed using  some caching,
ads you suggested, i have some code that i am going to put online soon
but i found that using caching  could render the database corrupted on
same use case,  so i need to discuss this  changes before proposing to
be merged.

Bye!
C.

bastibe:
Looks good to me. Good idea, too!

cage2:
Hi @bastibe !

> Looks good to me. Good idea, too!

Thank you, :) i am actually a big fan of magic number for file formats! :)
Maybe we could extend this types of test for other file formats in the future, we'll see! 

Bye!
C.

PR #98, opened 2021-02-05
Fixed 'annotate-previous-annotation-ends'

cage2:
Hi @bastibe !

'annotate-previous-annotation-ends' returned the wrong results when is testing an annotation that starts at position equals to `(pos-min).

I usually paste here the commit message (reworded, maybe) but in this case the commit message is ridiculously long so i do not think this would be a good idea this time! :-D :-D

Bye!
C.

### DIFF
--- a/Changelog
+++ b/Changelog
@@ -1,3 +1,76 @@
+2021-02-05  cage
+
+        * annotate.el:
+
+	- fixed 'annotate-previous-annotation-ends'
+
+	'annotate-previous-annotation-ends'  scans   the  buffer  contents
+	backward looking for an annotation before a buffer position
+        (let's' call this position 'POS').
+
+	If an annotation exists on 'POS' the backward search starts from a
+	position less than the starts of said annotation.
+
+	But unfortunately in the local function
+	'annotate-previous-annotation-ends' the test to stop the searching
+	is wrong.
+
+	This is the code for the local function:
+
+	----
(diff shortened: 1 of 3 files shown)


bastibe:
Maybe I am misunderstanding this issue, but why is it a `/=` instead of a `>=`? It seems strange to me that we would need an explicit predicate for checking whether `start` equals `(pos-min)`. But, again, I am probably not understanding the issue, so consider this comment not as criticism, and merge when you're ready!

cage2:
On Wed, Feb 10, 2021 at 12:42:17AM -0800, Bastian Bechtold wrote:

Hi Bastian!

> Maybe I am misunderstanding this issue, but why is it a `/=` instead
> of a  `>=`? It seems  strange to me that  we would need  an explicit
> predicate  for checking  whether  `start`  equals `(pos-min)`.  But,
> again, I am  probably not understanding the issue,  so consider this
> comment not as criticism, and merge when you're ready!

You are  correct! In  fact the  original test was  wrong and  this the
patch substitutes `/=` with `>'. :)

I checked again the changes after  your message because i was not sure
if the test  should be '>' or  '>=', fortunately seems to  me that the
behaviour of  the function was  the same regardless of  the comparison
predicate  used; good  that your  message  made me  rethink about  the
modification! :)

Bye!
C.

PR #102, opened 2021-04-16
Prevented asking for annotation when trying to annotate a newline character

### LINKED ISSUE #100: Annotation without region/word
Thank you for this library, it is very useful.

Is it expected that when I call `annotate-annotate` at end of line or on an empty line, the annotation doesn't get created at all? It is a bit surprising because the prompt shows normally. I would ideally like for the annotation to just be anchored to the end of line.

Note that annotating the beginning-of-line works for me: the annotation anchors to the first space I guess.

cage2:
Hi!!!

When an user try to annotate an newline (something is not possible to have because of the way the library is designed) the prompt asking for annotation text is shown anyway, even no annotation appears, i found this confusing at best.

I removed the prompt from popup in this case.

Fixes #100 

Bye!!
C.

### DIFF
--- a/NEWS.org
+++ b/NEWS.org
@@ -198,7 +198,7 @@
   Fixed highlight color  of annotated text that starts  from the first
   character of the buffer's content.
 
-- 2021-02-05 V1.1.5 Bastian Bechtold, cage ::
+- 2021-03-17 V1.1.5 Bastian Bechtold, cage ::
 
   Removed compilation  warnings, one of  the problem highlighted  by a
   warning was actually preventing this package working on Doom Emacs.
@@ -208,3 +208,21 @@
 
   Many thanks  to many people  that helped discovering  and suggesting
   fix for these problems!
+
+- 2021-04-27 V1.2 Bastian Bechtold, cage ::
+
+  New feature. When the customizable variable
+  `annotate-endline-annotate-whole-line' is not nil (default t), and
+  and the user try to annotate a newline the whole line is annotated
+  instead (or the next if the line is empty).
+
+  If the line contains a single annotation that cover all the line
+  annotating the newline Will ask to edit the annotation. If
+  `annotate-endline-annotate-whole-line' is nil annotating a newline
+  will signal an error.
+
+  Also this version generates smaller database as files left with no
+  annotations will not be saved on disk.
+
+  Finally   annotating   read-only    buffers   (especially   deleting
+  annotations) should works without problems.
--- a/annotate.el
+++ b/annotate.el
@@ -198,6 +198,13 @@ has been modified outside Emacs."
   :type 'boolean
   :group 'annotate)
 
+(defcustom annotate-endline-annotate-whole-line t
+ "Whether trying to annotate the end of line character will
+annotate the whole line before (or after if the line is composed
+by the newline character only) instead."
+  :type 'boolean
+  :group 'annotate)
+
 (defconst annotate-prop-chain-position
   'position)
 
@@ -586,7 +593,13 @@ specified by `from' and `to'."
                     (when (annotate-annotation-at (1- chain-end))
                       (annotate--cut-left-annotation last-of-chain-to-cut)))
                   (when delete-enclosed
-                    (annotate-delete-chains-in-region chain-end region-stop)))))
+                    (annotate-delete-chains-in-region chain-end region-stop))))
+              (annotate-line (eol)
+                (let* ((bol (annotate-beginning-of-line-pos)))
+                  (goto-char bol)
+                  (set-mark (point))
+                  (goto-char eol)
+                  (annotate-annotate))))
     (let ((annotation (annotate-annotation-at (point))))
       (cond
        ((use-region-p)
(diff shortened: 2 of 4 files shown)


indigoviolet (approved):
looks good to me. Thank you!

bastibe:
Looks good. It might be useful to flash a `(message` though, describing that you can't annotate newlines.

cage2:
On Sat, Apr 17, 2021 at 08:29:56AM -0700, Bastian Bechtold wrote:

Hi!

> Looks good. It might be useful to flash a `(message` though, describing that you can't annotate newlines.

I agree is useful  (and not too difficult too ;-)), and  i am going to
add a message in the PR i am preparing!

Bye!
C.

cage2:
Hi!!

I tried to implement what @indigoviolet  suggested in #100 , hope it works fine! :)

Bye!
C.

indigoviolet (commented):
Hi @cage2,

Thank you for working on this feature. I pulled your branch and tested it locally:

- annotating by grabbing the whole line works great!
- However, when I try to edit an existing "endline" annotation, with my cursor on the end of line, it tries to create a new annotation instead of grabbing the old annotation (ie. the prompt does not show my old annotation text). If I move my cursor over the line and then try to edit, things work as expected.
- A minor point: the message `The end of line can not be annotated` does not respect the `annotate-use-messages` setting, not sure this was deliberate.

cage2:
Hi everyone!

@indigoviolet  thank for testing this patch you too! 

I have changed the behaviour of the software that automatically switch to updating an existing annotation on the line that the user is annotating using the newline as "anchor". But this happens only if the annotation on the line is the only one present, in case there was more than a annotation instead of choosing in an arbitrary way which one gets updated i preferred to owerwrite the whole line instead.

Regarding the error message `annotate-use-messages` i intended it as a variable to switch off and on only informative messages not errors.

Bye and thank you again!!
C.

annotate.el
```diff
@@ -595,11 +595,23 @@ specified by `from' and `to'."
                   (when delete-enclosed
                     (annotate-delete-chains-in-region chain-end region-stop))))
               (annotate-line (eol)
-                (let* ((bol (annotate-beginning-of-line-pos)))
-                  (goto-char bol)
-                  (set-mark (point))
-                  (goto-char eol)
-                  (annotate-annotate))))
+                (let* ((bol                     (annotate-beginning-of-line-pos))
```
bastibe, line 640: These sub-functions are easy enough to follow, but on the whole `annotate-annotate` is becoming a bit too long for my liking. A few comments would improve readability a lot, I feel, or even breaking up into multiple functions.

bastibe:
If I had a line like this:

    this is an [annotated] line

with `[]` being annotated, there are now three behaviors:

- point on `[annotated]` changes the annotation
- point on any other word adds an annotation to that word
- point at EOL changes `[annotated]`?

I feel the last one is weird, since it does not change an annotation under point, as the other ones do.

I have a feeling that annotating end-of-line should only change whole-line annotations, not *any* annotation on the line.

What do you think?

annotate.el
cage2: Hi @bastibe !!!

You are right, there is no reason (except laziness ;-) ) that prevents adding documentation to these functions. As far as i know local functions does not supports docstring so i have added documentation as mere comments. I hope this is OK.

Bye!
C.

cage2:
> If I had a line like this:
> 
> ```
> this is an [annotated] line
> ```
> 
> with `[]` being annotated, there are now three behaviors:
> 
>     * point on `[annotated]` changes the annotation
> 
>     * point on any other word adds an annotation to that word
> 
>     * point at EOL changes `[annotated]`?

Yes, this is the behaviour i tried to code :-)
> 
> I feel the last one is weird, since it does not change an annotation under point, as the other ones do.
> 
> I have a feeling that annotating end-of-line should only change whole-line annotations, not _any_ annotation on the line.
> 
> What do you think?

In principle I liked the idea to update the annotation on the line if it i was the only one on the line, then i tried to do as you suggested and i feel more intuitive this behaviour of the command! :)

So this is the way annotation newline works at the moment (pasted from new documentation ;-)):

``` lisp
          
;; if  the line  contains no  text before  the newline
;; annotate the next line with text, if any
;;
;; if the line contains a single annotation that spans
;; the whole line update the existing annotation
;;
;; if  the line  contains no  annotation [or more than one annotation, this is missing from the documentation :-(] 
;; annotate  the
;; whole line that terminate at `eol'

```
If 

I hope i get your suggestion right :)

Bye!
C.

bastibe:
Looks good to me! Great work as always!

cage2:
Hi @bastibe !

Very well! I have changed the documentation and the NEWS and Changelog files, as usual.
I have marked this version as 1.2.0, as there is a new feature, not mere bugfix, i hope is OK. :)

Also, given problems we had in the past i checked bytecode compilation, and gives no warning fortunately! :)

I propose waiting some more hours if @indigoviolet has considerations to write and then merging!

Bye!!
C.

indigoviolet:
@cage2 - just had one comment: https://github.com/bastibe/annotate.el/issues/104#issuecomment-827805461

NEWS.org
```diff
@@ -198,7 +198,7 @@
   Fixed highlight color  of annotated text that starts  from the first
   character of the buffer's content.
 
-- 2021-02-05 V1.1.5 Bastian Bechtold, cage ::
+- 2021-03-17 V1.1.5 Bastian Bechtold, cage ::
```
bastibe, line 201: Why was the date changed?

cage2: Hi!

I made an error updating NEWS file from 1.1.4 to 1.1.5, i copied the entry from the former to write the latter but i forgot to change the release date that was, in fact, duplicated (you can check the file NEWS on the master branch, 1.1.4 and 1.1.5 share the same release date). I was able to find a plausible release date from commit log messages and used to fix the 1.1.5 entry.

Bye!
C.

Issue #86, opened 2020-11-24
Annotate Mode breaks `comment-dwim`


gopar:
When selecting a region (only works with selecting one line) in emacs and calling `comment-dwim`, an error happens in which the selected region is not commented out and ends up getting indented instead.

When  annotate mode is off, problem doesn't happen

The following stacktrace is an example of the error I get

```
Debugger entered--Lisp error: (args-out-of-range 511 512)
  remove-text-properties(511 512 (display nil))
  (save-excursion (goto-char end) (end-of-line) (remove-text-properties (point) (1+ (point)) '(display nil)))
  (let ((saved-undo-list (copy-tree buffer-undo-list t))) (buffer-disable-undo) (save-excursion (goto-char end) (end-of-line) (remove-text-properties (point) (1+ (point)) '(display nil))) (setq buffer-undo-list saved-undo-list) (buffer-enable-undo))
  (progn (setq inhibit-modification-hooks t) (let ((saved-undo-list (copy-tree buffer-undo-list t))) (buffer-disable-undo) (save-excursion (goto-char end) (end-of-line) (remove-text-properties (point) (1+ (point)) '(display nil))) (setq buffer-undo-list saved-undo-list) (buffer-enable-undo)))
  (unwind-protect (progn (setq inhibit-modification-hooks t) (let ((saved-undo-list (copy-tree buffer-undo-list t))) (buffer-disable-undo) (save-excursion (goto-char end) (end-of-line) (remove-text-properties (point) (1+ (point)) '(display nil))) (setq buffer-undo-list saved-undo-list) (buffer-enable-undo))) (setq inhibit-modification-hooks g1))
  (let ((g1 inhibit-modification-hooks)) (unwind-protect (progn (setq inhibit-modification-hooks t) (let ((saved-undo-list (copy-tree buffer-undo-list t))) (buffer-disable-undo) (save-excursion (goto-char end) (end-of-line) (remove-text-properties (point) (1+ (point)) '(display nil))) (setq buffer-undo-list saved-undo-list) (buffer-enable-undo))) (setq inhibit-modification-hooks g1)))
[...]

PR #106, opened 2021-04-30
Prevented calling 'annotate--remove-annotation-property' on a narrowed buffer

### LINKED ISSUE #103: 'annotate--remove-annotation-property' results in wrong behavior of comment-region
Hi,

There is an issue that a function 'annotate--remove-annotation-property' results in 'args-out-of' error when doing comment-or-uncomment-region as following:

**Back trace**

```
Debugger entered--Lisp error: (args-out-of-range 866 867)
  remove-text-properties(866 867 (display nil))
  annotate--remove-annotation-property(831 834)
  comment-region-internal(830 863 "/* " " */" nil nil nil nil t)
  comment-region-default-1(830 863 nil)
  #f(compiled-function () #<bytecode 0x1ff6f5aab7dd>)()
  combine-change-calls-1(830 864 #f(compiled-function () #<bytecode 0x1ff6f5aab7dd>))
  comment-region-default(830 863 nil)
  comment-region(830 863 nil)
  comment-or-uncomment-region(830 863)
  crux-duplicate-and-comment-current-line-or-region(1)
  funcall-interactively(crux-duplicate-and-comment-current-line-or-region 1)
  #<subr call-interactively>(crux-duplicate-and-comment-current-line-or-region t nil)
  elmacro-record-command(#<subr call-interactively> crux-duplicate-and-comment-current-line-or-region nil nil)
  apply(elmacro-record-command #<subr call-interactively> (crux-duplicate-and-comment-current-line-or-region nil nil))
  call-interactively(crux-duplicate-and-comment-current-line-or-region nil nil)
  command-execute(crux-duplicate-and-comment-current-line-or-region)
```

**Generating the problem**

In the following code (c-mode), when I try 'comment-or-uncomment-region', its behavior is abnormal.

error:
```
static int uvc_ioctl_ctrl_map(struct uvc_video_chain *chain,
	struct uvc_xu_control_mapping *xmap)
{
	/* struct uvc_control_mapping *map;
	struct uvc_control_mapping *map;
	unsigned int size;
```

expect:
```
static int uvc_ioctl_ctrl_map(struct uvc_video_chain *chain,
	struct uvc_xu_control_mapping *xmap)
{
	/* struct uvc_control_mapping *map; */
	struct uvc_control_mapping *map;
	unsigned int size;

```

When I remove a bunch of code in the function as below, the issue has gone but I don't know whether this is okay.

```
(defun annotate--remove-annotation-property (begin end)
  "Cleans up annotation properties associated with a region."
  (when (> (buffer-size)
           0)
    (annotate-with-inhibit-modification-hooks
     ;; copy undo list
     (let ((saved-undo-list (copy-tree buffer-undo-list t)))
       ;; inhibit property removal to the undo list (and empty it too)
       (buffer-disable-undo)
       ;; restore undo list
       (setf buffer-undo-list saved-undo-list)
       (buffer-enable-undo)))))
```

cage2:
Hi @bastibe !

From the commit message:

``For  some reason  (perhaps related  to font-locking?)   comment-region
command  was calling  this  function,  the fact  is  that the  command
narrows the  buffer before  actually adding comments,  so the  call to
`point'  inside  the annotate--remove-annotation-property's  body  may
fall  outside of  the buffer  raising  an error.  This patch  prevents
calling this function on narrowed buffer.''

This patch should fix #103 and i feel #86 too.

Bye!
C.

### DIFF
--- a/Changelog
+++ b/Changelog
@@ -1,3 +1,17 @@
+021-04-30  cage
+
+        * annotate.el:
+
+	- prevented calling 'annotate--remove-annotation-property' on a
+	narrowed buffer.
+
+	For some reason (perhaps related to font-locking?) comment-region
+	command was calling this function, the fact is that the command
+	narrows the buffer before actually adding comments, so the call to
+	`point' inside the annotate--remove-annotation-property's body may
+	fall outside of the buffer raising an error. This patch prevents
+	calling this function on narrowed buffer.
+
 2021-04-23  cage
 
         * annotate.el:
--- a/NEWS.org
+++ b/NEWS.org
@@ -226,3 +226,8 @@
 
   Finally   annotating   read-only    buffers   (especially   deleting
   annotations) should works without problems.
+
+- 2021-05-05 V1.2.1 Bastian Bechtold, cage ::
+
+  This version fixes a bug that prevented command like
+  'comment-region' to works properly when annotate-mode was active.
--- a/annotate.el
+++ b/annotate.el
@@ -1228,8 +1228,8 @@ a        a**
 
 (defun annotate--remove-annotation-property (begin end)
   "Cleans up annotation properties associated with a region."
-  (when (> (buffer-size)
-           0)
+  (when (and (> (buffer-size) 0)
+             (not (buffer-narrowed-p)))
     (annotate-with-inhibit-modification-hooks
      (annotate-with-disable-read-only
       ;; copy undo list
@@ -1242,8 +1242,7 @@ a        a**
           ;; annotated newline used to be
           (end-of-line)
           ;; strip dangling display property
-          (remove-text-properties
-           (point) (1+ (point)) '(display nil)))
+          (remove-text-properties (point) (1+ (point)) '(display nil)))
         ;; restore undo list
         (setf buffer-undo-list saved-undo-list)
         (buffer-enable-undo))))))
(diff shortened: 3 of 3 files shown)


bastibe:
Looks good to me!

The endless intricacies of Emacs continue to amaze. I wonder if anyone knows all the myriad ways in which Emacs can show the contents of a text file.

cage2:
Hi @bastibe !

> The endless intricacies of Emacs continue to amaze. 

I could not agree more! :) I still wonder if the patch proposed is the best solution to the issue reported as there are so many variable on the table! Fortunately Emacs seems to have all the functions in its standard library to fix things, at least so far! ;-D :D

I have updated the docs and the version number (1.2.1), i am going to merge this patch in the next hours!

Bye and thanks for checking the code! :)
C.

cage2:
OK, merging! :)

bastibe:
By the way, feel free to just write your name in the changelog. It'd be entirely justified!

cage2:
Hi!!

> By the way, feel free to just write your name in the changelog. It'd be entirely justified!

I am actually happy to put both names: yours and mine in the changelog
as you  reviewed each single pull  request i made (except  the simpler
ones, maybe)  in more  than a  year, each review  helped to  shape the
program as much as  the code i wrote, in my opinion.  In short i think
you deserve the credits :)

Anyway if  this makes you uncomfortable  (for any reasons, or  even no
reasons :))  i would not  add your name  in the future,  totally your
choice and i will respect that! :)

Bye!
C.

Issue #104, opened 2021-04-20
Should annotations files with no annotations be saved?


indigoviolet:
I noticed the [source code](https://github.com/bastibe/annotate.el/blob/54ac759facadacbfea5c1e7c2975e2da6434cdda/annotate.el#L1317) appears to prevent this, but I see files containing no annotations like this

```
(("~/dev/retinanet-examples/csrc/cuda/utils.h" nil "5d6550d74a31d9b6987c158b8773dca3"))
```
Note that I have customized `annotate-file` to be local to the file. Is this a bug?

cage2:
Hi!

Very good catch! Fixed in #102 (or so i hope!). ;-)

Bye!
C.

indigoviolet:
@cage2 - Thank you for your work on this! I just pulled #102 and tested this, and now I see that the annotation file is still saved, but it contains the text `nil`.

cage2:
Hi @indigoviolet !

I have checked the issue but the last version of the package seems not to be affected by this problem, may i ask to recheck using this use case?

1. set the  database annotation file targeting another file, or just make a backup of the db

``` lisp 
(setq annotate-file "/a/file")
```
2. annotate a file

3. save the annotated file

4.  check that the annotation database contains an entry like:

``` lisp
(("~/an/annotated/file" ((90530 90548 "annotation text" "annotated-text")) "hex-digits"))
```
5. delete the annotation from the annotated file
6.  save again the (no more) annotated file
7.  the annotation database file should contains just `nil`

Thank you!  
C.

indigoviolet:
@cage2 That is the behavior I see as well. 

But I think it is a bit inconvenient even to have `nil` in the annotation database file, because it means that if I open a file with `annotate-mode` on, then saving that file will lead to an empty annotation database file even if I never added any annotations.

So there are two cases:

1. I never added any annotations and saved a buffer with `annotate-mode` on -> ideally no database file should be created
2. I removed all annotations from a buffer that previous had them, and the annotation database file was previously created -> in this case I don't feel that strongly but it would be nice to remove the ("empty" == `nil` containing) annotation database file.  

Anyway, I'm also fine with shipping #102 as is for now - you have improved a lot of things in that PR! This is not such a big deal for me.

cage2:
Hi!

> @cage2 That is the behavior I see as well.

Sorry i did not read your previous message properly! :(

> But I  think it  is a  bit inconvenient  even to  have `nil`  in the
[...]

PR #107, opened 2021-05-07
Prevented saving a file with an empty database;

cage2:
Hi!

This patch try to address #104.

The general idea is (in function `annotate-dump-annotation-data`) to prevent the annotation file to be saved, or delete it, if the annotation database is empty (i.e. there are not buffers with annotated text when  `annotate-dump-annotation-data` is called).

Bye!
C.

### DIFF
--- a/annotate.el
+++ b/annotate.el
@@ -7,7 +7,7 @@
 ;; Maintainer: Bastian Bechtold
 ;; URL: https://github.com/bastibe/annotate.el
 ;; Created: 2015-06-10
-;; Version: 1.2.1
+;; Version: 1.3.0
 
 ;; This file is NOT part of GNU Emacs.
 
@@ -58,7 +58,7 @@
 ;;;###autoload
 (defgroup annotate nil
   "Annotate files without changing them."
-  :version "1.2.1"
+  :version "1.3.0"
   :group 'text)
 
 ;;;###autoload
--- a/annotate.el
+++ b/annotate.el
@@ -155,6 +155,13 @@ database is not filtered at all."
   :type 'boolean
   :group 'annotate)
 
+(defcustom annotate-database-confirm-deletion t
+ "If non nil a prompt asking confirmation before deleting a
+database file that is going to be empty after saving an annotated
+file will be shown"
+  :type 'boolean
+  :group 'annotate)
+
 (defcustom annotate-annotation-max-size-not-place-new-line 15
  "The maximum `string-width` allowed for an annotation to be
  placed on the right margin of the window instead of its own line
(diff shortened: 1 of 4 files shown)


annotate.el
```diff
@@ -1565,19 +1572,32 @@ annotation."
...
+           (delete-confirmed-p (or (not annotate-database-confirm-deletion)
+                                   (string= (read-from-minibuffer (format confirm-message
+                                                                           annotate-file))
+                                            "y"))))
+      (if delete-confirmed-p
+          (condition-case err
+              (delete-file annotate-file t)
+            (error (message "error removing annotation database: %S"
+                            (error-message-string err))))
+        (annotate-dump-annotation-data data t)))))
```
bastibe, line 1600: Why the additional recursion here (and the new optional argument)? I'm curious: did you run into a case where you were able to write data to the file, but not delete it?

bastibe:
Looks reasonable.

cage2:
Hi!

> Why the additional  recursion here (and the  new optional argument)?
> I'm curious: did  you run into a  case where you were  able to write
> data to the file, but not delete it?

The  recursive  call  is  made  when  the  user  totally  deleted  the
annotations in all  their files but asked not to  delete the database,
in this case an empty (contains only  'nil') is saved, this is the old
behaviour actually.

Or at least, this are the intentions of the code i wrote ;-) but seems
to work in my testing.

Bye!
C.

cage2:
Hi!

> Looks reasonable.

Good! Honestly i did not like the idea to delete entirely the file but
the motivations  suggested in the report  was valid to me  so changing
the  managing of  empty database  was  the right  thing to  do, in  my
opinion; i am  happy to know that  you also think the  changes this PR
introduce are good! Now I would  wait some times (a week?) waiting for
feedback from  the user that  opened the issue report  before merging
this, do you think is it OK?

Bye!
C.

indigoviolet:
@cage2 Thank you for your work on this change. I pulled and tested it and it seems to work perfectly!

cage2:
Hi!

> @cage2 Thank you for your work on this change. I pulled and tested it and it seems to work perfectly!

That's wonderful! I need to change the accessory files and version and i am going to merge this PR.

Bye and thanks for your help and suggestions!
C.

cage2:
Hi @bastibe !

I have updated version to 1.3.0 as there is a non trivial changes in the program behaivour here.
Also i have added just me to the NEWS file, but if you want to be added tour name too, you know i am totally OK with that!

Bye!
C.

annotate.el
cage2: Hi!

The  recursive  call  is  made  when  the  user  totally  deleted  the
annotations in all  their files but asked not to  delete the database file,
in this case an empty (contains only  'nil') db is saved, this is the old
behaviour actually.

Or at least, this are the intentions of the code i wrote ;-) but seems
to work in my testing.

Bye!
C.

cage2: Hi!

The  recursive  call  is  made  when  the  user  totally  deleted  the
annotations in all  their files but asked not to  delete the database file,
in this case an empty (contains only  'nil') db is saved, this is the old
behaviour actually.

Or at least, this are the intentions of the code i wrote ;-) but seems
to work in my testing.

Bye!
C.

bastibe:
Great work, as always!

cage2:
Hi @bastibe !

Thank you!! :)

Merging in a sec! :)

Bye!
C.

Issue #108, opened 2021-05-25
[Replace] button does not edit annotations; throws abbreviate-file-name: Wrong type argument


glvno:
Hello again!

After reinstalling Spacemacs, I'm suddenly having trouble with the replace button. It fails to edit the annotation and instead throws `abbreviate-file-name: Wrong type argument: stringp, nil`. It seems possible that this has something to do with the latest [merge](https://github.com/bastibe/annotate.el/pull/107), as abbreviate-file-name is only invoked once, and in a section of lisp that got heavily revised.

Another new and potentially related issue is that the \*annotations* buffer occasionally fails to display the correct annotated text. So the database has the correct region defined numerically as well as the correct string, the text in foo.org is correctly highlighted, but what should be displayed in \*annotations* is offset by an arbitrary number of characters, usually between 2 and 10.

So if foo.org contains the line

> foo **bar baz**

with the section in bold denoting annotated text, annotate-show-annotation-summary will generate an annotations buffer that includes something like

> ** Annotated text: "oo bar b"

where the annotated text in quotes is offset by two or more characters.

Please let me know if you need further clarification or information. Also, let me know if I should split the second problem into a separate issue.

Thanks so much for your work on this great package!

cage2:
Hi @glvno !

Just a quick reply to write that the first issue is confirmed and reproducible on my system.
I can not reproduce the second but just because i need some time to investigate it more to understand what is happening.
In the meanwhile i would start working on the edit issue.

Thanks a lot!
Bye!
C.

glvno:
Thank you! For future reference, here's a tidy screenshot of the offset issue:

![image](https://user-images.githubusercontent.com/25870908/119578811-d57bf180-bd82-11eb-9b30-6ab14acf2ba3.png)

cage2:
Hi @glvno !

Turned out that the "[replace]" button's callback was severely broken! I just pushed a patch (#109) that supposed to fix this issue.

May i ask to take a look at it?

Thanks!
C.

glvno:
Hi @cage2,

Sorry for the delay;  I don't have a lot of experiences with quelpa recipes, so it took me a second to figure out how to test your patch.

The patch works wonderfully. Thanks so much for your quick work!

I use this package extensively, and I cannot overstate how grateful I am for your help.

I'm going to keep an eye out to see if i can figure out how to consistently replicate the offset issue.

Thanks again!

m

glvno:
Hello again!

[...]

PR #109, opened 2021-05-26
Fixed callback for "[replace]" button in summary window

cage2:
Hi! 

The  callback function  was neither  properly saving  the annotation  database nor refreshing the buffer's annotations.

This is a preliminary works to address the first issue reported in #108 .

Bye!
C.

### DIFF
--- a/annotate.el
+++ b/annotate.el
@@ -2470,6 +2470,20 @@ results can be filtered with a simple query language: see
                                     (info-setup filename (current-buffer))
                                     (buffer-substring-no-properties annotation-begin
                                                                     annotation-end)))
+              (build-snippet-from-buffer (filename annotation-begin annotation-end)
+                (let ((visited-buffer (find-buffer-visiting filename)))
+                  (when visited-buffer ;; a buffer is visiting the file
+                    (with-current-buffer visited-buffer
+                      (let ((raw-snippet (buffer-substring-no-properties annotation-begin
+                                                                         annotation-end)))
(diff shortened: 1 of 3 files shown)


cage2:
Hi!

This PR should fix the bugs reported in #108 !

Bye!
C.

annotate.el
```diff
@@ -2389,6 +2389,24 @@ pressed."
         (annotate-update-visited-buffer-maybe filename)
         (annotate-show-annotation-summary query nil nil)))))
 
+(cl-defun annotate-wrap-text (text &optional (wrapper "\""))
+  (concat wrapper text wrapper))
+
+(cl-defun annotate-unwrap-text (text &optional (wrapper "\""))
```
bastibe: I wonder if a solution using `substring` wouldn't be cleaner. Regular expressions seem a bit heavy handed for this job.

bastibe:
Another set of great bugfixes! These execution order bugs are always tricky to figure out.

Looks good to me!

annotate.el

I agree that using regex is too much and, moreover, makes the code a bit ugly (in my opinion); i pushed a version that uses substring.

Bye!
C.

cage2:
Hi @bastibe !

Thanks for reviewing this PR! I would wait a few of days for feedback from the user who reported #110 and then i would merge this anyway (after changing ~version number and~ docs, as usual, of course! :-)).

Bye!!
C.

PR #113, opened 2021-08-24
Added a command to delete an annotation

cage2:
Hi @bastibe !

I have added a new command `C-c C-d` to delete the annotation under the point in a buffer. This is a small patch as the code to build the command was more or less already there! :)

Sorry, i made a mistake and i am using the master branch instead of a custom branch for this pull request. :(

Bye!
C.

### DIFF
--- a/annotate.el
+++ b/annotate.el
@@ -2102,7 +2102,8 @@ This function is not part of the public API."
 (defun annotate--confirm-annotation-delete ()
   "Prompt user for delete confirmation.
 This function is not part of the public API."
-  (funcall annotate-y-or-n-prompt-function annotate-confirm-deleting-annotation-prompt))
+  (or (not annotate-annotation-confirm-deletion)
+      (funcall annotate-y-or-n-prompt-function annotate-confirm-deleting-annotation-prompt)))
 
 (cl-defun annotate-delete-annotation (&optional (point (point)))
   "Command  to  delete  an  annotation,  `point'  is  the  buffer
--- a/annotate.el
+++ b/annotate.el
@@ -2108,7 +2108,7 @@ point)."
   (when-let ((annotation (annotate-annotation-at point)))
     (let* ((delete-confirmed-p (annotate--confirm-annotation-delete)))
       (when delete-confirmed-p
-        (annotate--delete-annotation-prevent-modification annotation)))))
+        (annotate--delete-annotation-chain-prevent-modification annotation)))))
 
 (defun annotate-change-annotation (pos)
   "Change annotation at point. If empty, delete annotation."
@@ -2128,7 +2128,7 @@ point)."
          ((string= "" annotation-text)
           (let* ((delete-confirmed-p (annotate--confirm-annotation-delete)))
             (when delete-confirmed-p
-              (annotate--delete-annotation-prevent-modification highlight))))
+              (annotate--delete-annotation-chain-prevent-modification highlight))))
          ;; annotation was changed:
          (t
           (change highlight)))))))
--- a/annotate.el
+++ b/annotate.el
@@ -1995,9 +1995,7 @@ This function is not part of the public API."
    (save-excursion
      (with-current-buffer (current-buffer)
        (let* ((chain         (annotate-find-chain annotation))
-              (filename      (annotate-actual-file-name))
-              (info-format-p (eq (annotate-guess-file-format filename)
-                                 :info)))
+              (filename      (annotate-actual-file-name)))
          (dolist (single-element chain)
            (goto-char (overlay-end single-element))
            (move-end-of-line nil)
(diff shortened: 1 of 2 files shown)


annotate.el
```diff
@@ -2076,15 +2082,26 @@ from a chain where `annotation' belong."
        (t
         (move-overlay last-annotation last-annotation-starting-pos new-ending-pos))))))
 
-(defun annotate--delete-annotation-prevent-modification (annotation)
+(defun annotate--delete-annotation-chain-prevent-modification (annotation)
+"Delete an annotation chain backing up and restoring modification
+status of the buffer before deltion occured."
```
bastibe: typo: "deletion"

annotate.el
```diff
@@ -2084,22 +2084,29 @@ from a chain where `annotation' belong."
...
 "Delete an annotation chain backing up and restoring modification
-status of the buffer before deltion occured."
+status of the buffer before deletion occured.
+
+This function is not part of the public API."
   (annotate-ensure-annotation (annotation)
     (annotate-with-restore-modified-bit
      (annotate--delete-annotation-chain annotation))))
 
+(defun annotate--confirm-annotation-delete ()
```
bastibe, line 2097: I think we should use `yes-or-no-p` instead of implementing the prompt ourselves. That way the form of the prompt is standardized. For example, I like to alias my `yes-or-no-p` to `y-or-n-p`, to save me some typing.

bastibe (approved):
Another great addition, thank you!

I am not quite sure whether it is a good idea to prompt the user when deleting annotations. It is certainly something I would disable immediately. But if you find it useful, that's reason enough.

annotate.el
cage2: Hi @bastibe !

> Another great addition, thank you!

Thanks!

> I am not quite sure whether it is a good idea to prompt the user when deleting annotations. It is certainly something I would disable immediately. But if you find it useful, that's reason enough.

The funny thing is that i find asking any time annoying too! :-D, i am very happy to switch the default behaviour to not ask for confirmation! :-)

Bye!
C.

annotate.el
cage2: Hi!

I just did not know about this function, sure is the best solution instead of a custom one! (and also i learnt something new today)!

Bye!
C.

cage2:
Hi!

I forgot to actually use the variable `annotate-annotation-confirm-deletion` 😓! I think everything is ready for merging! :)

Bye!
C.

annotate.el
```diff
@@ -219,6 +219,11 @@ by the newline character only) instead."
   :type 'boolean
   :group 'annotate)
 
+(defcustom annotate-y-or-n-prompt-function 'yes-or-no-p
```
bastibe: I think the `defcustom` is actually unnecessary. But it doesn't hurt, either.

annotate.el
```diff
@@ -164,7 +164,7 @@ file will be shown."
   :type 'boolean
   :group 'annotate)
 
-(defcustom annotate-annotation-confirm-deletion t
+(defcustom annotate-annotation-confirm-deletion nil
```
bastibe, line 167: I wonder if it is even necessary to implement deletion-confirmation at all if we're disabling it by default anyway. I'll leave that up to you, though.

bastibe:
Hi @cage2,

looks good to me! I have added two similar comments, regarding the prompt. Feel free to leave all customization in there, or delete them completely. I can see useful arguments for both paths.

It might be that I am in a deletion-mood, since I have spent the last few weeks excising dead code at work. So my perception might be skewed.

Otherwise, merge when you're ready!

cage2:
Hi @bastibe !

Sorry for the late reply!

Yes, very likely the use of `annotate-y-or-n-prompt-function` could be considered a case study for over engineering! :-D
Instead i would prefer to keep the boolean variable to asking for confirmation before deleting an annotation as some users could be more comfortable to keep a "safety net" for annotations (even if they are ephemeral by nature).

Bye!
C.

bastibe:
I must say that if there is one big "lesson" I have learned in my programming, it is to avoid unnecessary complication. Too many times have I seen student projects crumble under the weight of their abstractions. Too often have I seen projects reduce to something much simpler once I removed all the "architecture". I now try to code stuff in the most straight-forward manner first, and only introduce abstractions and indirections when they are absolutely provably necessary. Or at least that's what I try to do.

Anyway, I don't mean this as criticism to this pull request. Emacs is very forgiving with regards to configuration variables, and the prompt is a reasonable and unproblematic customization option. Merge when you're ready!

cage2:
Hi @bastibe !

In retrospective i think i have added the customizable variable `y-or-n-p` just because of my ignorance about elisp, i did not know about `defalias` for example. Now i wonder if there could be something superfluous hidden in the code that i could remove using simpler or already existing constructs in the standard library. 

Anyway, let's merge this, for now! :)

Bye!
C.

PR #115, opened 2021-09-30
Fix delete ghost db

cage2:
Hi @bastibe !

When an user has no annotation database file on the filesystem and no buffer annotated, the program keeps asking to delete the file that does not exists.

I never noted that bug before because my annotation database is never empty! :-D

Bye!
C.

### DIFF
--- a/Changelog
+++ b/Changelog
@@ -1,3 +1,14 @@
+2021-09-30 cage
+
+        * NEWS.org,
+        * annotate.el:
+
+        - increased version number.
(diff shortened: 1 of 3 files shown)


bastibe:
Great find! Always difficult to find bugs in use cases you don't have.

cage2:
Hi @bastibe !

In fact i only stepped in this bug by chance! :) 

OK, merging!

Bye!
C.

PR #128, opened 2022-04-08
Increased version number.

cage2:
Hi!

This version also reverts a form and updates NEWS file (also Changelog needs to be updated as well).

Bye!
C.


cage2:
OK, i think everithing should be fine now!

Sorry again!
C.

Issue #123, opened 2022-02-06
[new feature] using margins to print annotation's text


cage2:
Emacs have, indeed, margins as display property:

https://www.gnu.org/software/emacs/manual/html_node/elisp/Display-Margins.html

Annotations could be displayed there.

Problems:
- horizontal position (left? Right? Depending of the language?)
- vertical position (on the first line of the annotated text or on the last?)
- replace current way of display annotations? Or add this new way of displaying as optional, an in that case, what default?
- ...

Comments welcome! :)
Bye!
C.

bastibe:
That looks useful indeed! If I remember correctly, however, the margins are a fixed portion of the buffer width. I thought they were mostly used for single-character "annotations", like the little arrow that says a line continues off screen. 

So if we enlarge to the margins to print into them, this might restrict the buffer width, and cut off long lines. I might be wrong, though.

The advantage of our current approach is that we can have a long, non-intrusive annotation at the end of a short line, and overflow gracefully for long lines. This flexibility would probably be lost if we print to the margins.

But at any rate, there's nothing wrong with providing both display methods as an option. Perhaps even chosen automatically based on the frame width (and content width?).

cage2:
Hi!!

> That looks useful indeed!

I think the same!

> If I remember correctly, however, the margins are a fixed portion of
> the   buffer  width.   I   thought  they   were   mostly  used   for
> single-character "annotations",  like the  little arrow that  says a
> line continues off screen.

I was confused at  first (i was remembering the same as  you) but if I
have read correctly the documentation what you are mentioning here are
the fringe indicators:

https://www.gnu.org/software/emacs/manual/html_node/elisp/Fringe-Indicators.html

While the margin are also an unreachable portion of the buffer but can contains arbitrary text (again if i did not checked wrong the documentation).

https://www.gnu.org/software/emacs/manual/html_node/elisp/Display-Margins.html

>  So if  we enlarge  to the  margins to print  into them,  this might
> restrict the buffer width, and cut off long lines. I might be wrong,
> though.

I  quite sure  you are  right, I  think  we have  to try  and see  the
results,  i  am  thinking  making  this  way  to  display  annotations
optional, but who knows? Maybe we  could like so much the margins that
this could be the default in the future, we'll see!

Bye!
C.

bastibe:
I think there was some discussion about customizable display styles at some point. At the time we were musing different styles such as
- annotations after the end of the annotated line
- annotations always on the next (virtual) line
[...]

PR #129, opened 2022-04-23
Fix resize hook

cage2:
Hi!

`Window-size-change-functions` wants a list of functions that accepts a frame as an argument, whilst
we was using 'font-lock-flush' that takes two integer arguments.  The results was that the annotation was not properly resized.

This patch fix this problem.

Bye!
C.

### DIFF
--- a/Changelog
+++ b/Changelog
@@ -1,3 +1,13 @@
+2022-04-23 cage
+
+        * annotate.el:
+
+        - fixed  function that  is  called on  'window-size-change-functions',
+        this function is  called with a single frame as  an argument, whilst
+        we was using 'font-lock-flush' that takes two integer arguments.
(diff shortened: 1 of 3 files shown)


bastibe (approved):
Looks good!

Another bug fixed. It seems that the bugs you fix are becoming more and more specific, which is cool to see! That's a good proxy that annotate is getting closer to being bug-free 🎉.

cage2:
Hi @bastibe!

> Looks good!

Thanks! :-)

> Another bug fixed. It seems that the bugs you fix are becoming more and more specific, which is cool to see! That's a good proxy that annotate is getting closer to being bug-free tada.

But I expect that, when i will start working on  #123, an entirely new breed of bugs will arise! 😊

Bye an thanks again for reviewing the code!
C.

PR #130, opened 2022-05-26
Added procedures to generate and use a single database for each annotated file.

### LINKED ISSUE #124: [new feature] Using a file 'FILENAME.notes' to hold the annotations of file 'FILENAME'
This is a big change, as means get rid of the database.

Comments welcome!
Bye!
C.

cage2:
Hi!

This is my attempt to close #124. The patch is simple. So simple that i think would be a good idea to test it for a week before merging, so some possible regressions can be highlighted. ;-D

Bye!
C.

### DIFF
--- a/Changelog
+++ b/Changelog
@@ -1,3 +1,10 @@
+2022-05-26 cage
+
+        * annotate.el:
+
+        - added procedures to generate and use a single database for each
+        annotated file.
+
 2022-04-23 cage
 
         * annotate.el:
--- a/NEWS.org
+++ b/NEWS.org
@@ -1,4 +1,10 @@
-- 2022-04-23 v1.5.4 cage::
+- 2022-05-26 v1.6.0 cage ::
+
+  This version  allows users  to instruct the  program to  generate an
+  annotation  database for  each annotated  file, instead  of using  a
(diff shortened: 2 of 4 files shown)


bastibe:
Simple and beautiful! I like it!

cage2:
Hi @bastibe 

> Simple and beautiful! I like it!

Great! I am going to update the necessary documentation files and will merge that!

My only doubt is: this new version will be the 1.6.0, right? 😅

Bye!
C.

bastibe:
> this new version will be the 1.6.0, right? 😅

I would say, so, yes. A new feature means a new minor version, if we're sticking with semantic versioning.

cage2:
On Thu, Jun 02, 2022 at 04:43:18AM -0700, Bastian Bechtold wrote:
> > this new version will be the 1.6.0, right? 😅
>
> I would say, so, yes. A new feature means a new minor version, if we're sticking with semantic versioning.

I  like semantic  version, in  fact i  try to  stick to  this kind  of
numbering in  nearly all my projects,  but I am also  always unsure to
decide the next version number. And that is why I ask so often.  :) :)

Bye!
C.

PR #132, opened 2022-06-23
- removed the  use of  'after-save-hook', used  'kill-buffer-hook' and 'kill-emacs-hook', instead.

cage2:
Using these  two hooks prevents  the buffer  to be marked  as modified
when,   in  fact,   only   the  annotation   has   been  changed   (or
added/deleted).

Bye!
C.

### DIFF
--- a/annotate.el
+++ b/annotate.el
@@ -7,7 +7,7 @@
 ;; Maintainer: Bastian Bechtold <bastibe.dev@mailbox.org>, cage <cage-dev@twistfold.it>
 ;; URL: https://github.com/bastibe/annotate.el
 ;; Created: 2015-06-10
-;; Version: 1.6.0
+;; Version: 1.6.1
 
 ;; This file is NOT part of GNU Emacs.
 
@@ -58,7 +58,7 @@
 ;;;###autoload
 (defgroup annotate nil
   "Annotate files without changing them."
-  :version "1.6.0"
+  :version "1.6.1"
   :group 'text)
 
 (defvar annotate-mode-map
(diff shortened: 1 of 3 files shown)


annotate.el
```diff
@@ -58,7 +58,7 @@
 ;;;###autoload
 (defgroup annotate nil
   "Annotate files without changing them."
-  :version "1.6.0"
+  :version "1.6.1"
```
bastibe: This is technically a regression, i.e. a major version change. But that feels silly. It should probably be a minor version change, though, as it is more than a bugfix, right?

cage2:
On Thu, Jun 30, 2022 at 11:39:44PM -0700, you wrote:
> @bastibe commented on this pull request.
>

Hi!

>
> > @@ -58,7 +58,7 @@
>  ;;;###autoload
>  (defgroup annotate nil
>    "Annotate files without changing them."
> -  :version "1.6.0"
> +  :version "1.6.1"

>  This is technically a regression,  i.e. a major version change. But
> that  feels silly.  It should  probably be  a minor  version change,
> though, as it is more than a bugfix, right?


Thanks for  discussing this,  i was  not sure  the new  version number
attribution, in  fact. In  a sense  this patch fixes  a tiny  bug, but
doing that changes dramatically the  behaviour of the program. What is
your opinion if we would move to 1.7.0?

Bye!
C.

bastibe:
I think 1.7.0 is the right thing to do.

cage2:
Hi @bastibe !

> I think 1.7.0 is the right thing to do.

Very well! I am going to change the relevant files and will merge! :)

Thanks!
C.

PR #133, opened 2022-08-02
Fix run hook when emacs killed

cage2:
Hi!

There is a bug in the previous version of the code that  prevented the database to be updated with the annotations contained in non visible buffers, when Emacs is closed. Those annotations was, in effect, lost.

Although i am not sure the code is optimized this patch remove the bug described above.

Sorry for the inconvenience.
C.

### DIFF
--- a/Changelog
+++ b/Changelog
@@ -1,3 +1,15 @@
+2022-08-02 cage
+
+        * NEWS.org,
+        * annotate.el:
+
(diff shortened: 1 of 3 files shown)


cage2:
Hi!

I a bit busy but i hope to have some moments to check again this patch and merge in the next hours.

Sorry for the delay.
C.

Issue #136, opened 2022-09-03
It would be nice if "annotate-use-echo-area" worked without the use of the mouse


alienbogart:
For example, leaving the cursor above the note would display its content in the echo area instead of requiring me to use the mouse. I don't really use the mouse on Emacs so it would be nice to be able to do this with the keyboard. Thanks!

cage2:
Hi @alienbogart !

I like your idea! The only issue is that i do not know how to make Emacs generates (and reacts) to an event like "the cursor is placed over some annotated text", three possible approaches maybe could be explored (not an exhaustive list below, these are just the ones coming to mind):

1. using an hooks, but i do not think an hook run when the cursor is moved exists (but i may be wrong)
    - will running this hooks kill the performance?
2. using `run-with-idle-timer` with an high frequency
    - even if it does works, will the users be annoyed with the delay associated with the timer?
    - again, what about the performance?
3. Transform the annotated text in a button 
    - will necessarily require the user to press the button to popup the annotation's text?
    - the appearance of the annotate text will be affected by the fact that ha been transformed in a button? I think yes, but how and how much?

I think # 2 is the best option, so i would starts exploring that. If you have some other idea let me know, of course! :)

Bye!
C.

cage2:
Hi @alienbogart !

I have added a few lines of code in this branch:

https://github.com/cage2/annotate.el/tree/print-message-annotate-under-point

The code will print the annotation under cursor in the minibuffer, if the text is annotated of course!

This branch is no more that a proof of concept, nevertheless i would be glad if you would be try it and tell me if the results of this patch are more or less the whet you expected from your suggestion.

Thanks  in advance!
C.

cage2:
Hi @alienbogart !

I believe that the patch #137  implement your suggestion. Let me know it it does not!

Thanks again!
C.

PR #137, opened 2022-09-21
Print message annotate under point

cage2:
Hi!

With this patch i believe #136 could be closed!

Bye!
C.

### DIFF
--- a/Changelog
+++ b/Changelog
@@ -1,11 +1,37 @@
+2022-09-21 cage
+
+        * NEWS.org,
+        * README.org,
+        * annotate.el:
+
+        - added a customizable  variable to switch on and off  the printing of
+        annotations in the minibuffer;
+        - updated version number;
+        - updated NEWS.org;
+        - updated README.org.
+
+2022-09-16 cage
+
+
+        Merge pull request #135 from cage2/fix-closing-indirect-buffer
+        Merge branch 'master' into print-message-annotate-under-point
+
+2022-09-08 cage
+
+        * annotate.el:
+
+        - added idle timer to print annotation under point.
+
 2022-08-31 cage
 
+        * Changelog,
         * NEWS.org,
         * annotate.el:
 
         - updated, and added if missing, docstrings.
         - updated NEWS.org.
         - removed a duplicated label using a constant.
+        - updated changelog.
 
 2022-08-25 cage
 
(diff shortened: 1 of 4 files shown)


bastibe:
Neat! Great idea!

I wonder about the default delay of 0.1 s, though. That seems unexpectedly low, I'd have expected something like 0.5 s, as with e.g. completion libraries. On the other hand, maybe you'd really want the annotation to show up immediately, i.e. with a very short delay. I guess that means the delay should be configurable.

cage2:
On Thu, Sep 22, 2022 at 01:27:59AM -0700, you wrote:

Hi Bastibe!!

> Neat! Great idea!

All the merits goes to alienbogart, they suggested that feature.

> I  wonder about  the  default delay  of 0.1  s,  though. That  seems
> unexpectedly low,  I'd have expected  something like 0.5 s,  as with
> e.g. completion  libraries. On  the other  hand, maybe  you'd really
> want the annotation  to show up immediately, i.e. with  a very short
> delay. I guess that means the delay should be configurable.

Honestly there is no reason that  makes me choose that value as delay;
it is more or  less a random value but, as you  wrote, small enough to
be unnoticeable.  I think your suggestion to make it customizable (and
defaulting to  0.5s) is the right  thing to do! Thanks  for suggesting
that change! :)

Bye!
C.

cage2:
Hi @bastibe!

Finally i have found the time to (hopefully!) complete this patch! :) :)

Bye!
C.

bastibe:
Looks good!

cage2:
Hi @bastibe !

Very well, i am going to merge this in a minute!

I am sorry i am slowing down working on this package but i have been very busy lately, but i can guarantee i still want to take care of this package! :)

Thanks for reviewing my code!
C.

bastibe:
Do not worry, @cage2. There's absolutely no obligation to work on this. Please only work on annotate if and when it's fun to you.

cage2:
Hi @bastibe !

> Please only work on annotate if and when it's fun to you.

I find absolutely fun to work on annotate! :) 

I am sorry that lately i had not so much time to work on all my FLOSS projects. I hope things will get better in future, and also that new ideas for this package comes from us or the people that are using the package (like in this case)!

Bye!
C.

PR #138, opened 2022-10-20
Fix integrate inplace

cage2:
Hi!

Unfortunately the command to integrate the annotations into the buffer's content was broken. :/ 
 This patch should fix the issue.

Bye!
C.


cage2:
Hi @bastibe !

A small patch to fix (hopefully!) a big issue! :D
Merging in a second.

Bye!
C.

PR #144, opened 2022-12-15
Prevented warning when saving a non annotated buffer

cage2:
Hi @bastibe !

A little patch to remove an annoying warning. :)

From NEWS.org

``` text
  This versions prevented printing of a warning

  `annotate-save-annotations' printed a warning  if the buffer was not
  based on a file's contents. But  the warning was printed even if the
  buffer contained no annotations at all.

  This  version prints  the warning  about  missing file  only if  the
  buffer is annotated (i.e. contains at least an annotation).
```

Bye!
C.

### DIFF
--- a/Changelog
+++ b/Changelog
@@ -1,21 +1,42 @@
+2022-12-15 cage
+
+        * annotate.el:
+
+        - prevented printing of warning when saving a buffer that has no file
+          under the hood and no annotation.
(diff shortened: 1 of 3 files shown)


bastibe:
Looks good! Another small usability improvement! Wonderful!

cage2:
Hi @bastibe !

Thanks for the review! I am merging in a second. :smile: 

Season greetings! :christmas_tree: 
C.

PR #146, opened 2023-02-02
Fix annotate shutdown

cage2:
Hi @bastibe !

I hope you are doing well!

I just discovered these errors that prevented a clean exit when disabling `annotate-mode` in a buffer, i was a bit sad but i think I have fixed the issue with this patch so I am fine again! :smile: :smile: 

Bye!
C.

### DIFF
--- a/Changelog
+++ b/Changelog
@@ -1,3 +1,20 @@
+2023-02-02 cage
+
+        * annotate.el:
+
+        - fixed limit for deleting text-property (the annotation's text).
+        - prevented capure of a variable in macro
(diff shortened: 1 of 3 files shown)


bastibe:
Tricky bug! Great work finding and fixing it!

cage2:
Hi @bastibe !

> Tricky bug! Great work finding and fixing it!

Thanks!

I think there is room to clean the changes i made with this patch a bit. But for now is better to have a working package, i guess. :) 

Also this reminds me i need to work on a group regression tests... :sweat: 

Bye!
C.

Issue #149, opened 2023-02-12
is it possble to hide annotation but keep showing annotate-highlght


emacs25:
for some reason,i dont want anybody to see my annotations but still keep annotate-highlight,how should i do?

closing annotate-mode disables annotaton but the highlight is disabled too.

cage2:
On Sat, Feb 11, 2023 at 04:58:49PM -0800, emacs25 wrote:

Hi!

> for some reason,i dont want anybody to see my annotations but still keep annotate-highlight,how should i do?
>
> closing annotate-mode disables annotaton but the highlight is disabled too.

This looks like definitely a good idea to me!

Bye!
C.

cage2:
Hi @emacs25 !

I have found some free time to start working on this feature! :)

I wrote a couple of function that perhaps can be useful to what you proposed here.

To get the new experimental feature just clone this fork of the package: 

https://github.com/cage2/annotate.el/

``` shell
$ git clone https://github.com/cage2/annotate.el/
$ git switch hide-annotation-text
```

somewhere in your init file just load the file `annotate.el` from this fork.

``` elisp
 (load "path/to/cloned/annotate-package/annotate.el")
```

then you will be able to launch a new command `annotate-toggle-annotation-text`.

Let me know if it is OK to you.
Bye!
C.

cage2:
Hi @emacs25 !

I think the patch #150 implements whet you asked in this thread. It is now possible to hide all the annotation's' text in a buffer or choose to hide the text for the annotation under the cursor.

Bye!
C.

emacs-newer:
sorry for replying so lately, something is wrong with my old account,and i create a new one

i will test it as soon as possible and give feedback.

thanks again

cage2:
On Thu, Feb 23, 2023 at 10:27:38PM -0800, you wrote:

Hi!

> sorry for replying so lately, something is wrong with my old account,and i create a new one
>
> i will test it as soon as possible and give feedback.

No problem! Take your time and thanks for testing!
Bye!
C.

emacs-newer:
i test for both english and CJK characters and they all work as expected.

just a problem maybe:

it will wait for a little time  when the text content is too big with a lot of annotations, escpecially for CJK characters.

is there anyway to improve the performance?

cage2:
On Fri, Feb 24, 2023 at 04:18:20PM -0800, you wrote:

Hi!

> i test for both english and CJK characters and they all work as expected.

Good!

[...]

Issue #147, opened 2023-02-09
how to loop color face when creating annotate?


emacs25:
great package which helps me a lot to read book
 with thinking.                                
i want more colors to show so i add annotation-
highlight-third....and annotate-annotation-thir
d...and so on.

please tell me how to modify "annotate--annotat
ion-builder" to enable them?

emacs25:
by the way,is it possible to adjust emacs theme with both light and dark?

cage2:
Hi @emacs25 !

The number of colors for the annotation is deeply ingrained in the code, unfortunately. Your proposed changes needs a lot of work and to be honest i do not know when i am going to find spare time to work on this improvements. If you want to file a patch i can support you writing it, if help is needed. Otherwise we have to wait that I finish working on the issue's backlog (and not only for this project!). The idea is nice, though! :smile_cat: 

Bye!
C.

cage2:
> by the way,is it possible to adjust emacs theme with both light and dark?

You mean adapt the annotation theme with the emacs theme? Likely yes, do you have some documents about how light and dark theme for works in emacs?

Thanks!
C.

emacs25:
maybe i should wait for your plan because i can't write elisp code for the function now.

i customize annotate color for different colors and add them into dark/light theme to adapt emacs theme.

cage2:
On Thu, Feb 09, 2023 at 01:58:57PM -0800, you wrote:

Hi!

> maybe i should  wait for your plan because i  can't write elisp code
> for the function now.

Thanks for your patience!

In the  meanwhile we can  discuss the issue; what  about if I  add the
chance to set  a custom color for each annotation,  using a prefix for
the command `annotate-annotate'.

I am thinking to something like that:

- two  customizable  variable  holding  a  list  of  theme  color  for
  annotations and  annotation's text (similar  to `annotate-highlight'
  and `annotate-annotation')

annotate-annotation-color-highlights → '((:underline "coral") (:underline "khaki") ...)

annotate-annotation-colors           → '((:background "coral" :foreground "black" :inherit default) (:background "khaki" :foreground "black" :inherit default) ...)

this  way you  could add  a  prefix to  annotate-annotate, where  said
prefix would be an index to  an element of the aforementioned lists, so
typing:

C-u 1 annotate-annotate RET

would create (or even modify?)  an annotation using the second element
of the color list.

What do  you think about this  feature? Would this be  good enough for
you?

Bye!
C.

emacs25:
[...]

PR #150, opened 2023-02-22
Hide annotation text

cage2:
Hi!

This patch supposed to implements what is described in #149 .

Bye!
C.

### DIFF
--- a/Changelog
+++ b/Changelog
@@ -1,3 +1,25 @@
+2023-02-22 cage
+
+        * annotate.el:
+
+        - added command 'annotate-toggle-annotation-text';
+        - removed customizable variable 'annotate-hide-annotations-text'.
+        - wrapped raw chain manipulation with 'annotate-chain-last-ring' to
+        get the last segment of a chain.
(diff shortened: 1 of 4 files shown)


bastibe:
That's a unique use case for annotate. Very interesting.

Your implementation looks good.

cage2:
Hi Bastian!

> That's a unique use case for annotate. Very interesting.

I agree, it would be an unusual scenario for this feature to be useful, but sometimes, for example, a big annotation is pushing the rest of the text too far and hiding it could be a temporary solution.
 
> Your implementation looks good.

Thanks! I would wait to merge this until @emacs-newer test this patch, they offered to check if i have implemented correctly the  feature they proposed. 

Bye!
C.

cage2:
Hi!

I would merge this because i have some ideas for #147 and i would like to work on that issue, that seems a lot interesting to me. :smile: 

Bye!
C.

cage2:
Hi!

OK, merging!  :rocket: :smile: 

Bye!
C.

PR #159, opened 2023-10-10
Saving color of each annotation in the database

cage2:
Hi @bastibe !

I thought, given that an user can specify the color theme for each annotation, could be a good idea to save this information in the database, making the change persistent so that can be restored later. The database format has changed but the change is backward compatible (but i will recheck tomorrow, just in case! :smile: ).

Bye!
C.

### DIFF
--- a/Changelog
+++ b/Changelog
@@ -1,19 +1,52 @@
-2023-09-25 cage <cage>
+2023-10-10 cage
 
-	* annotate.el:
+        * NEWS.org,
+        * annotate.el:
 
-	- fixed docstring.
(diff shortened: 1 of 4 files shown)


cage2:
Hi @bastibe !

I would like to merge this patch, as I think it adds an useful feature; what is your idea about merging this?

Bye!
C.

bastibe:
Looks good, as far as I can tell! Merge away!

cage2:
Hi @bastibe !

I know i can always count on your help! Thanks!

Merging in a sec!

Bye!
C.

PR #162, opened 2024-05-08
Fixed wrong coloration for adjacent annotations

cage2:
Hi @bastibe !

A minor (but annoying) bug fixed  (I fear won't be the last one! :smile: :smile:). 

Bye!
C.

----------------

Example:

``` text
  foo-bar
   ⏶  ^^^ ← annotated text
   |
   `cursor position
```
annotating  "foo-"  ("C-c  C-a")  would  set a  color  for  the  new annotation  equals to  the one  on the  left, the  correct behavior
would be to set a different color.

### DIFF
--- a/Changelog
+++ b/Changelog
@@ -1,5 +1,42 @@
+2024-05-08 cage
+
+	* annotate.el:
+
+	- increased version number.
+
+2024-04-17 cage
+
+	* annotate.el:
+
+	- fixed color's annotation when the new annotated text end border lays
+	just next to an already annotated text
+	for example:
+	foo-bar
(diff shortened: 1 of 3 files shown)


bastibe:
Quite the subtle bug you found there!

Your fix look good!

cage2:
Hi @bastibe !

> Quite the subtle bug you found there!

Yes, the number of possible configuration text/ position of annotation seems quite high, i do not believe i still stepped into all of them! :smile: 
 
> Your fix look good!

Thanks! I am merging in a minute!
C.

PR #163, opened 2024-10-10
Prevented annotating text on the last line with no newline

cage2:
Hi @bastibe !

I do not usually allow a file to terminate without a newline character, but sometimes I make a mistake and, this time, i discovered this bug! :smiley: 

In previous version annotating the last "line" of a buffer that does not ends with a newline just underlined the annotated text without showing the annotation text; now an error is printed on the minibuffer, instead.

Bye!  
C.

### DIFF
--- a/annotate.el
+++ b/annotate.el
@@ -46,7 +46,7 @@
 ;; the previous annotation.  Use M-x annotate-export-annotations to
 ;; save annotations as a no-difference diff file.
 
-;; Important note: annotation can not overlaps and newline character
+;; Important note for developers: annotation can not overlaps and newline character
 ;; can not be annotated.
 
 ;;; Code:
@@ -358,6 +358,10 @@ in the customizable colors lists:
 
 (define-error 'annotate-empty-annotation-text-error
   "Empty annotation text"
+  'annotate-error)
+
+(define-error 'annotate-no-new-line-at-end-file-error
+  "No newline found at the end of the buffer"
   'annotate-error)
 
 (define-error 'annotate-db-file-not-found
@@ -732,6 +736,8 @@ and
                   (let ((annotation-text (read-from-minibuffer annotate-annotation-prompt)))
                     (condition-case nil
                         (annotate-create-annotation start end annotation-text nil color-index)
+                      (annotate-no-new-line-at-end-file-error
+                       (user-error "Missing newline at the end of the buffer"))
                       (annotate-empty-annotation-text-error
                        (user-error "Annotation text is empty"))))))
               (cut-right (region-beg region-stop &optional delete-enclosed)
@@ -1117,8 +1123,8 @@ An example might look like this:"
   "Export all annotations as a unified diff file.
 An example might look like this:
 
---- .../annotate.el/annotate.el	2015-06-19 15:13:36.718796738 +0200
-+++ .../annotate.el/annotate.el	2015-06-19 15:13:36.718796738 +0200
+--- .../annotate.el/annotate.el 2015-06-19 15:13:36.718796738 +0200
++++ .../annotate.el/annotate.el 2015-06-19 15:13:36.718796738 +0200
 @@ -73,5 +73,5 @@
  ;;;###autoload
  (defface annotate-highlight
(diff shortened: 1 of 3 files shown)


annotate.el
```diff
@@ -1,13 +1,13 @@
 ;;; annotate.el --- annotate files without changing them  -*- lexical-binding: t; -*-
-;; Copyright (C) 2015 Bastian Bechtold and contributors:
+;; Copyright © 2015 Bastian Bechtold and contributors:
```
bastibe, line 2: 😍

bastibe:
If I'm reading this correctly, we're throwing an error if the file does not end in a newline, right? Does this prevent the creation of an annotation on just the last line, or anywhere in the file?

I agree with you that files which don't end in a newline are an abomination, but on the other hand, I have seen plenty of file formats that require the most bizarre things imaginable. Some poor soul may be required to end files without a newline. We probably shouldn't punish them any further.

annotate.el
cage2: The magic of UTF-8! :smile:

cage2:
Hi @bastibe !

> If I'm reading this correctly, we're throwing an error if the file does not end in a newline, right? 

Exactly!

> Does this prevent the creation of an annotation on just the last line, or anywhere in the file?

I thought the same in principle, but I tested the changes and turned out that, fortunately, it does not!

>
> I agree with you that files which don't end in a newline are an abomination, but on the other hand, I have seen plenty of file formats that require the most bizarre things imaginable. Some poor soul may be required to end files without a newline. We probably shouldn't punish them any further.

Your observation makes sense and I agree that the best solution (maybe the proper one) would be to just annotate the last line even if it is not exactly a line. But the fact is that we store the annotation text sticking it as a property of the newline character, if the newline does not exists the annotation text just won't show up (but it will if the user just hit "enter" key at the end of the buffer, and also will if an alternative way to show the annotation is used, like using the minibuffer).

So i can see just four scenarios:

1. we leave the things as is, that is, if the last improper line is annotated just the annotated text is underlined and the annotation text will be shown only if a newline character is added to the buffer, likely would be a nice idea to add a warning to inform the user about this problem;

2. show an error and refuse to annotate (essentially what this patch does);

3. figure out another way to show annotation on the buffer different to add property to the newline character an makes the annotation works anywhere;

4. any other idea I am not able to figure out.

Looking forward to your opinion!
C.

EDIT: Thinking about it, I prefer the first solution.

bastibe:
I agree with you. Solution 1 is probably best.

But I wonder: we already insert line breaks into the annotations, when rendering them onto multiple lines. Could we do that for and EOF/no-NL annotation as well, and just add a fake newline to the overlay? I guess we'd have to switch to overlaying the final character of a line, instead of the newline.

cage2:
On 12/10/24 10:49, you wrote:

Hi Bastibe!

> I agree with you. Solution 1 is probably best.
>
> But I wonder: we already insert line breaks into the annotations, when rendering them onto multiple lines. Could we do that for and EOF/no-NL annotation as well, and just add a fake newline to the overlay? I guess we'd have to switch to overlaying the final character of a line, instead of the newline.

Yes, if we would just be able to add some kind of "ephemeral", so to say, newline character, maybe this would be the perfect solution. I will try if i am going to be able to accomplish this, otherwise i will implement the first idea of the list.

In the meanwhile thanks for taking some time of a Saturday to answer to my message, I wish you an happy rest of weekend, bastibe! 🙂
C.

bastibe:
> In the meanwhile thanks for taking some time of a Saturday to answer to my message, I wish you an happy rest of weekend, bastibe! 🙂

Same to you, cage!

cage2:
Hi @bastibe !

with the latest commits I think I have essentially implemented the first item in the list above:
 
> we leave the things as is, that is, if the last improper line is annotated just the annotated text is underlined and the annotation text will be shown only if a newline character is added to the buffer, likely would be a nice idea to add a warning to inform the user about this problem;

I just added the chance, for the user, to "fix" the buffer (appending a newline) and retry to annotate again the buffer.

To be honest I am not fully satisfied by this patch, but I failed to implement the "ghost newline", so to say; I tried using the text's "display" property but, even if the new line is present in the buffer seems to be not recognized by our code.

I would take a week of time to try to figure out something and, if I will not, i would rather merge this, what do you think?

Bye!
C.

bastibe:
The current solution looks fine to me. We're communicating clearly that there is a problem, offer a solution to fix it, and don't degrade too much if we can't. That's much more reasonable than most software! Great idea with the prompt!

cage2:
Hi @bastibe !

Your last message is reasonable and comforting! I am going to merge this patch in a second!

Thanks.  
C.

PR #165, opened 2025-02-16
[add] autosave option

### LINKED ISSUE #164: Autosave option
As far as I can see from the code, annotations are saved either manually with `annotate-save-annotations`, or automatically on `kill-emacs`. If I disable `annotation-mode` for a buffer before the annotations are saved, all the annotations for this buffer are lost.

So, if my Emacs instance is running for a long time (which is a usual situation), for all this time annotations are stored only in the variables inside annotated buffers. If Emacs would crash, all the data is lost.

What do you think about an autosave option, which would run `annotate-save-annotations` on each confirmed change of each annotation? I.e. when a user hits `C-c C-a`, `annotate-save-annotations` should be called after they hit `RET` in the minibuffer.

(Seems that it could be done with a small patch in `annotate-annotate` function: we just need to check an option in the very end of the function, and if it is non-nil, to call `annotate-save-annotations`.)

krvkir:
https://github.com/bastibe/annotate.el/issues/164

### DIFF
--- a/annotate.el
+++ b/annotate.el
@@ -251,7 +251,7 @@ of lines. The center of the region is the position of the
 annotation as defined in the database."
   :type 'number)
 
-(defcustom annotate-autosave t
+(defcustom annotate-autosave nil
   "Whether annotations should be saved after each user action,
 e.g. new annotation created, existing one amenede or deleted."
   :type 'boolean)
--- a/annotate.el
+++ b/annotate.el
@@ -253,7 +253,7 @@ annotation as defined in the database."
 
 (defcustom annotate-autosave nil
   "Whether annotations should be saved after each user action,
-e.g. new annotation created, existing one amenede or deleted."
+e.g. new annotation created, existing one amended or deleted."
   :type 'boolean)
 
 (defconst annotate-prop-chain-position
@@ -892,7 +892,8 @@ and
                         (forward-line 1)
                         (goto-char (annotate-end-of-line-pos))
                         (annotate-annotate))))))))))))
-      (if annotate-autosave (annotate-save-annotations)))))
+      (when annotate-autosave
+	(annotate-save-annotations)))))
 
 (defun annotate-toggle-annotation-text ()
   "Hide annotation's text at current cursor's point, if such annotation exists."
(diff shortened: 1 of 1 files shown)


annotate.el
```diff
@@ -251,6 +251,11 @@ of lines. The center of the region is the position of the
 annotation as defined in the database."
   :type 'number)
 
+(defcustom annotate-autosave t
+  "Whether annotations should be saved after each user action,
+e.g. new annotation created, existing one amenede or deleted."
```
cage2: there is a a typo here

annotate.el
```diff
@@ -886,7 +891,8 @@ and
                                             ; with proper text
                         (forward-line 1)
                         (goto-char (annotate-end-of-line-pos))
-                        (annotate-annotate)))))))))))))))
+                        (annotate-annotate))))))))))))
+      (if annotate-autosave (annotate-save-annotations)))))
```
cage2: usually when the `if` has a single branch the macro `when` is used instead

Also please break the form as:

``` lisp
(when  test
   forms)
```

annotate.el
```diff
@@ -2526,7 +2532,8 @@ point)."
     (let* ((delete-confirmed-p (annotate--confirm-annotation-delete)))
       (when delete-confirmed-p
         (annotate--delete-annotation-chain annotation)
-        (font-lock-flush)))))
+        (font-lock-flush)))
+    (if annotate-autosave (annotate-save-annotations))))
```
cage2: the same as above here

cage2:
Hi @krvkir !

The patch looks fine to me, thanks for accepting my suggestions.
Let's wait if @bastibe has comments and, if OK, i am going to merge your patch in the next hours.

Thank you for your work!
C.

cage2 (approved):

cage2:
Hi!

I think it is OK to take responsibility for merging this patch.

Thanks @krvkir for contributing to this package! I believe `annotate.el` could be the right project to starts with elisp as it is not a big program ma neither a trivial one too. So, if you want, feel free to contribute more!

I am not going to update the version number right after merging this patch, the number will be increase after the documentation files will be updated too. I will take care of that.

Bye and thanks again!
C.

Issue #167, opened 2025-03-09
Is there a way to start with all annotations hidden?


siovan:
Hi! 

Thank you for this wonderful library. It is exactly what I needed. Please forgive me if this is mentioned somewhere and I missed it, but is there an option to start with all the annotations hidden?

Thanks!

cage2:
Hi @siovan !

I am happy that you found this package useful!

>  is there an option to start with all the annotations hidden

That feature is not directly available from the package, but you can likely use an hooks in combination with the command `annotate-toggle-all-annotations-text`.

Try to evaluate that code:

``` lisp 
(add-hook 'annotate-mode-hook (lambda () (annotate-toggle-all-annotations-text)))
```

I tried on my system and it works, but be aware that the command above (`annotate-toggle-all-annotations-text`), is a bit slow, unfortunately. :(

Bye!
C.

siovan:
Thank you so much for the reply. The solution works perfectly.

PR #168, opened 2025-03-19
Optimize searching for annotations

cage2:
Hi!

the function `annotate-annotations-overlay-in-range` is sadly inefficient (see: https://github.com/bastibe/annotate.el/issues/167#issuecomment-2708983595), this patch optimize the scanning loop; for comparison, this is the time spent by Emacs calling `(annotate-toggle-all-annotations-text)` when performed on the `annotate.el` source code file with a single annotation:

``` text
Elapsed time: 5.043135s (3.782171s in 175 GCs)
```

This is the same function called with the patched code:

``` text
Elapsed time: 0.000276s
```

If I have not introduced new bugs :sweat_smile: , looks a good improvements.

Bye!
C.

### DIFF
--- a/Changelog
+++ b/Changelog
@@ -1,10 +1,30 @@
+2025-03-19 cage
+
+	* annotate.el:
+
+	- optimized 'annotate-annotations-overlay-in-range' by jumping from an
+	annotation to another, instead of scanning every single character of
+	the buffer.
+
+2025-03-06 cage
+
+
+	Merge pull request #166 from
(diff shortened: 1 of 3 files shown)


bastibe:
Looks great! That's a much more efficient loop, scanning the source text in O(n) instead of O(n²). Good catch!

cage2:
Hi @bastibe !

Shame on me for having written such an inefficient code! :smile: :sweat_smile: 

Bye!
C.

PR #169, opened 2025-04-30
Added annotation's text expansion feature

cage2:
Hi!

This PR add a new feature: a string matching a regular expression in the annotation's text may be replaced with the output of a customizable shell command, e.g.

``` lisp
(\"%d\" \"date +%Y-%m-%d\" t)
```

will expand the sub string `%d` with the results of running the command `date +%Y-%m-%d`, that is

``` text
"%d foo" → "2025-04-23 foo"
```

The last element on the list is a boolean that indicates if the command must be trimmed using `string-trim`.

Also, please note that i have flipped `annotate-autosave` from `nil` to `t`, because i found it a very useful feature! :smile: 

Bye!
C.

### DIFF
--- a/Changelog
+++ b/Changelog
@@ -1,10 +1,19 @@
+2025-03-20 cage
+
+	Merge pull request #168 from cage2/optimize-searching-for-annotations
+
 2025-03-19 cage
 
+	* Changelog,
+	* NEWS.org,
 	* annotate.el:
 
 	- optimized 'annotate-annotations-overlay-in-range' by jumping from an
 	annotation to another, instead of scanning every single character of
 	the buffer.
+	- increased version number;
+	- updated NEWS file and changelog;
+	- fixed docstring.
 
 2025-03-06 cage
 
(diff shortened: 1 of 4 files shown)


bastibe:
Wow, what a neat feature! This is happening at annotation time, right? So the string that's saved in the annotation, contains the expanded text.

cage2:
Hi @bastibe 

> This is happening at annotation time, right? So the string that's saved in the annotation, contains the expanded text.

Yes, the results of the shell command permanently replaces the placeholder in the annotation's text and the expanded text is saved in the database; to be honest, after your message, i am tempted to add expansions when annotation is printed too! :smile: :wink:  But i have to think if i would add this feature for the fun of implementing it, or it is actually a good idea! :smiley: 

Bye!
C.

cage2:
> But i have to think if i would add this feature for the fun of implementing it, or it is actually a good idea! 😃

In the meanwhile, I am merging this patch, by the way! :smile:

Issue #170, opened 2025-05-08
Compile warnings


jcs090218:
```
Compiling internal form(s) at Wed May  7 23:39:38 2025
Leaving directory ‘c:/Users/XXX/AppData/Roaming/.emacs.d/elpa/annotate-20250507.1212’

Compiling file c:/Users/XXX/AppData/Roaming/.emacs.d/elpa/annotate-20250507.1212/annotate.el at Wed May  7 23:39:38 2025
Entering directory ‘c:/Users/XXX/AppData/Roaming/.emacs.d/elpa/annotate-20250507.1212/’
annotate.el:103:10: Warning: in defcustom for ‘annotate-highlight-faces’:
    ‘list’ without arguments
annotate.el:109:10: Warning: in defcustom for
    ‘annotate-annotation-text-faces’: ‘list’ without arguments
annotate.el:260:12: Warning: custom-declare-variable
    ‘annotate-annotation-expansion-map’ docstring wider than 80 characters
annotate.el:260:12: Warning: custom-declare-variable
    ‘annotate-annotation-expansion-map’ docstring has wrong usage of unescaped
    single quotes (use \=' or different quoting such as `...')

In annotate-annotation-set-position:
annotate.el:479:2: Warning: docstring wider than 80 characters

In annotate-overlay-maybe-set-position:
annotate.el:487:2: Warning: docstring wider than 80 characters
annotate.el:487:2: Warning: docstring has wrong usage of unescaped single
    quotes (use \=' or different quoting such as `...')

In annotate-annotation-interval:
annotate.el:1697:2: Warning: docstring wider than 80 characters

In annotate-load-annotations:
annotate.el:1829:2: Warning: docstring wider than 80 characters

In annotate-create-annotation:
annotate.el:2242:2: Warning: docstring wider than 80 characters

In annotate-summary-show-annotation-button-pressed:
annotate.el:2839:9: Warning: ‘info-setup’ is an obsolete function (as of
    30.1); use ‘info-pop-to-buffer’ instead

In annotate-show-annotation-summary:
annotate.el:2999:20: Warning: ‘info-setup’ is an obsolete function (as of
    30.1); use ‘info-pop-to-buffer’ instead
```

cage2:
Hi @jcs090218 !

Thanks for reporting these warnings, I think i have addressed all of them; in the next hours i will provide a patch (I want to test the new, info related, functions before to send a PR).

Thanks!
C.

cage2:
[...]

Issue #142, opened 2022-11-22
Adding regression tests


cage2:
The package needs some tests to check for regressions.

Manny approaches can be taken to do testing (assertion, code that run only during compilation, libraries, etc), i think ERT is a good candidate.

Roadmap:

- [ ] starting studying the library;
- [ ] write the actual tests.

Bye!
C.

PR #172, opened 2025-05-10
ci: Add basic CI

jcs090218:
For https://github.com/bastibe/annotate.el/issues/170#issuecomment-2867960445.

See the test result in https://github.com/jcs-PR/annotate.el/actions/runs/14944767061?pr=1.

### DIFF
--- a/annotate.el
+++ b/annotate.el
@@ -2849,9 +2849,9 @@ sophisticated way than plain text."
   "Display Info node FILE-OR-NODE in BUFFER.
 
 Compatibility wrapper for the function `info-setup' and `info-pop-to-buffer'."
-  (if (version< "30" emacs-version)
-      (info-pop-to-buffer file-or-node buffer)
-    (with-no-warnings
+  (with-no-warnings
+    (if (version< "30" emacs-version)
+        (info-pop-to-buffer file-or-node buffer)
       (info-setup file-or-node buffer))))
 
 (defun annotate-summary-show-annotation-button-pressed (button)
--- a/annotate.el
+++ b/annotate.el
@@ -8,6 +8,7 @@
 ;; URL: https://github.com/bastibe/annotate.el
 ;; Created: 2015-06-10
 ;; Version: 2.4.1
+;; Package-Requires: ((emacs "30.1"))
 
 ;; This file is NOT part of GNU Emacs.
 
--- a/annotate.el
+++ b/annotate.el
@@ -601,7 +601,7 @@ See also the customizable variables: `annotate-echo-annotation-timer' and
 `annotate-print-annotation-under-cursor'."
   (with-current-buffer (current-buffer)
     (when annotate-mode
-      (when-let ((annotation (annotate-annotation-at (point))))
+      (when-let* ((annotation (annotate-annotation-at (point))))
         (message "%s%s"
                  annotate-print-annotation-under-cursor-prefix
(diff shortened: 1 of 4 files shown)


cage2:
Hi @jcs090218 !

Thanks for your patch and also thanks for fix all the `when-let` macro calls; i did not know it was deprecated. :sweat: 

To be honest i feel this line: 

>  Package-Requires: ((emacs "30.1"))

To be a strong requirement. Would it break backward compatibility? I mean, would this changes render the package uninstallable on Emacs version < 30.1? I think we need to discuss this change with @bastibe . 

Bye!
C.

jcs090218:
> Thanks for your patch and also thanks for fix all the when-let macro calls; i did not know it was deprecated. 😓

It's deprecated in the snapshot version. No worries! ;)

> To be a strong requirement. Would it break backward compatibility? I mean, would this changes render the package uninstallable on Emacs version < 30.1?

I've tested with Emacs 26.1 and later (see: https://github.com/jcs-PR/annotate.el/actions/runs/14944721529/job/41986571480), and it appears the package is already broken on versions earlier than 30.1 due to the following error:

```
In end of data:
annotate.el:3777:1:Error: the function ‘info-pop-to-buffer’ is not known to be
    defined.
```

annotate.el
```diff
@@ -2843,6 +2843,15 @@ sophisticated way than plain text."
...
   'help-echo "Click to replace annotation")
 
+(defun annotate-info-setup (file-or-node buffer)
+  "Display Info node FILE-OR-NODE in BUFFER.
+
+Compatibility wrapper for the function `info-setup' and `info-pop-to-buffer'."
+  (if (version< "30" emacs-version)
+      (info-pop-to-buffer file-or-node buffer)
+    (with-no-warnings
+      (info-setup file-or-node buffer))))
```
jcs090218, line 2855: I've fixed the compatibility issue here.

annotate.el
```diff
@@ -1546,20 +1546,20 @@ surrounded by `BEGIN' and `END'."
...
-		(setf counter chain-end)
-		(when (and (>= chain-start from-position)
-			   (<= chain-end   to-position))
-		  (cl-pushnew annotation annotations)))
-	    (throw 'scan-loop t)))))
+        (setf counter chain-end)
+        (when (and (>= chain-start from-position)
+               (<= chain-end   to-position))
+          (cl-pushnew annotation annotations)))
+        (throw 'scan-loop t)))))
```
jcs090218: It looks like there are some tabs here. 🤔

jcs090218:
It seems like the minimum version is `27.1` since the function `diff-buffers` is added in `27.1`. 🤔 I've lower the minimum version to `27.1`.

Edit: See the latest CI result in https://github.com/jcs-PR/annotate.el/actions/runs/14949424866?pr=1.

annotate.el
cage2: That's great!

cage2:
Hi @jcs090218 !

Thanks for all your work, especially for taking care about backward compatibility! Looks all fine with me. Let's wait some time for comments and then i think this patch could be merged. Incidentally this patch also, albeit partially, solves #142!

Thanks again!
C.

bastibe:
Thank you for helping out with the CI!

cage2:
OK, I am going to merge this patch, many thanks to you all for your contribution!
C.

PR #173, opened 2025-05-12
- fixed type definition of customiziable variables and docstrings

cage2:
This patch try to implements suggestions from Emacs maintainers.

In details:

- the type of the customizable variables related to face specify a list of plist but without imposing the number of element in such plist, as we can not know in advance its length.

- In docstrings,  only symbols names are wrapped in `…', variable names are just upcased.

Bye!
C.

### DIFF
--- a/annotate.el
+++ b/annotate.el
@@ -86,9 +86,9 @@ See https://github.com/bastibe/annotate.el/ for documentation."
   :type 'file)
 
 (defcustom annotate-file-buffer-local nil
-  "If non nil (default `NIL'), for each annotated file \"filename\", a database
+  "If non nil (default `nil'), for each annotated file \"filename\", a database
 \"filename.notes\", containing the annotations, is generated in the
-same directory that contains `filename'."
+same directory that contains \"filename\"."
   :type 'string)
 
 (defcustom annotate-buffer-local-database-extension "notes"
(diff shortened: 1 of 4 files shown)


annotate.el
```diff
@@ -458,71 +457,71 @@ position (so that it is unchanged after this function is called)."
...
+  "Get the last ring of CHAIN."
   (car (last chain)))
 
 (defun annotate--remap-chain-pos (annotations)
-  "Remap `ANNOTATIONS' as an annotation \"chain\".
+  "Remap ANNOTATIONS as an annotation \"chain\".
 
 An annotation is a collection of one or more overlays that
-contains the property `ANNOTATE-PROP-CHAIN-POSITION'.
+contains the property `annotate-prop-chain-position'.
```
bastibe, line 511: why aren't these upcased?

bastibe:
Fascinating. Are these formatting rules spelled out somewhere?

cage2:
Hi @bastibe !

I was wondering the same, then I found that the Emacs Elisp manual cover that topic too.

https://www.gnu.org/software/emacs/manual/html_node/elisp/Documentation-Tips.html

I wish i had read this appendix before! :sweat: 

Bye!
C.

annotate.el
cage2: As far as i have understood, upcased text is for function arguments only, symbols supposed to be downcased and wrapped.

I am going to recheck the code, because i fear i have made some mistakes.

Bye!
C.

cage2: looks like i did. :sweat_smile: 

Bye!
C.

bastibe:
Not gonna lie, these are some complex formatting rules. But if that's how it's done, that's how we should do it, too. 

Thank you for maintaining this project so diligently!

cage2:
Hi @bastibe 

> Not gonna lie, these are some complex formatting rules. 

I think the same; fortunately the command `checkdoc` is able to identify a few mistakes category.
 
> Thank you for maintaining this project so diligently!

I try to do my best, thanks! :smile: 

I am happily surprised that the Emacs maintainers actually review the code for package even in NonGNU ELPA. 

Bye!
C.

PR #174, opened 2025-10-31
abbreviated buffer's file name to allow annotating remote file using TRAMP mode

cage2:
Hi @bastibe !

I hope you are doing well!

I started using TRAMP mode lately and I noted that annotate-mode was unable to annotate remote files.

Fortunately seems that there was just an inconsistency between the file name saved on the database (that was abbreviated) and the one got from the buffer using `annotate-actual-file-name` (that was not abbreviated).

This patch try to fix this issue.

Bye!
C.

### DIFF
--- a/Changelog
+++ b/Changelog
@@ -1,8 +1,38 @@
-2025-05-14 cage
+2025-11-01 cage
 
 	* annotate.el:
 
+	- wrapped 'file-exists-p' with 'ignore-errors' to catch TRAMP
+	connection errors.
+	- fixed docstring for 'annotate-file-exists-p'.
+
+2025-10-31 cage
+
+	* annotate.el:
+
+	- abbreviated buffer's file name to allow annotating remote file using
+	TRAMP mode;
+	- fixed indentation and moved function definition.
+
+2025-05-15 cage
+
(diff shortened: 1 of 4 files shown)


bastibe:
Tramp is such a cool technology! But I suppose we're storing the annotations locally? How does it represent network paths?

Anyway, I'm just asking out of curiosity. The PR looks good, as always!

cage2:
Hi @bastibe !

> Tramp is such a cool technology! But I suppose we're storing the annotations locally? How does it represent network paths?

Happy to try to answer your question!

This is how an actual remote path is stored in the annotation database:

``` text
/ssh:cage@sveglia:~/lisp/ediaf/launch-ediaf-rpi.sh
``` 
Nice things is that `(file-exists-p "path")` returns `non-nil` even if `path` points to a remote file (unfortunately it does signal an error if TRAMP is unable to open a connection, this is the reason I wrapped the calls to `file-exists-p` with `ignore-errors`).

> 
> Anyway, I'm just asking out of curiosity. The PR looks good, as always!

Thanks, i am going to update the docs and then merging that patch!

Bye!
C.

PR #175, opened 2025-11-02
Changed 'ignore-errors' with 'with-demoted-errors'

cage2:
Following the suggestion of an Emacs maintainer, i used `with-demoted-errors` in `annotate-file-exists-p` because, this way, an error message is printed when an error occurs while testing for a file existence (even the error is still ignored).

### DIFF
--- a/Changelog
+++ b/Changelog
@@ -1,10 +1,22 @@
+2025-11-02 cage
+
+	* annotate.el:
+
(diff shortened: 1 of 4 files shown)


cage2:
Hi @bastibe !

Given that the patch is trivial, i think it is OK if i am going to merge it.

Bye!
C.

